PR #2, opened 2018-01-17
Add initial Dockerfile.

angusiguess:
Currently, this Dockerfile:

- Mounts the repo we run build from to boot-clj
- Pulls all deps that clj-docs needs to run analysis

Tested against a couple of jars to ensure docs generate properly.

As far as getting documents out of the container we can either mount a volume or call `docker cp` to copy them out.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,6 @@
+FROM adzerk/boot-clj:latest
+ADD ./ /boot-clj
+WORKDIR /boot-clj
+# Pull deps in the build phase, this takes awhile
+RUN boot show -d
+ENTRYPOINT ["/usr/bin/boot"]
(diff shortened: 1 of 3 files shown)


martinklepsch:
Thanks! The first PR :tada: I'm excited!

The `Dockerfile` looks good. Two small requests requests:

- Move all docker related stuff into `docker/`
- add a Readme to the `docker/` directory with some basic instruction how to run stuff and get data in and out of the container.

angusiguess:
Hey @martinklepsch, we should be documented now. There are other ways to store the file with docker that could be useful later, but they're minor variations on what we already have here!

martinklepsch:
Awesome, I'll merge this now and have a play with it tomorrow.

Also big congrats on being the first ever contributor! :tada:

angusiguess:
❤️ thanks so much!

PR #4, opened 2018-01-19
Add --rm switch to remove stopped containers

angusiguess:
The current suggested commands for running the container leave stopped containers, which can be solved by adding the `--rm` flag when we run the containers.

- [x] Add flag
- [x] Ensure mounted directory still works


angusiguess:
Also `-it` doesn't really do anything for us in this context, removing

martinklepsch:
Sweet!

PR #5, opened 2018-01-21
ADR-0007: Introduce DocBundles for Non-API Docs

martinklepsch:
I've been thinking about how to integrate non-API documentation into Grimoire for a while and this ADR is a proposed approach for doing so.

To a large degree this approach sidesteps Grimoire's hierarchy but I think there is a place for documentation that does not relate to a namespace or def specifically. Also I hope that a future integration of Examples and other things outlined in the [stacks](https://github.com/arrdem/stacks) project might fill this gap. This however, is still mostly unfamiliar to library authors and so I feel like we should pick them up from a place they know.

Link to [a rendered version of the current state of the ADR](https://github.com/martinklepsch/cljdoc/blob/adr/doc-bundles/doc/adr/0007-introduce-docbundles-for-non-api-docs.md).

### DIFF
--- a/doc/adr/0007-introduce-docbundles-for-non-api-docs.md
+++ b/doc/adr/0007-introduce-docbundles-for-non-api-docs.md
@@ -1,0 +1,150 @@
+# 7. Introduce DocBundles for Non-API Docs
+
+Date: 2018-01-21
+
+## Status
+
+Proposed
+
+## Context
+
+Non API documentation such as articles and guides are just as
+important—if not more important—as API documentation. We want to
+provide an easy way for projects to publish a collection of articles
+in a hierarchy that makes sense for the project.
+
+Yada does an [exemplary job](https://juxt.pro/yada/manual/index.html)
+at providing this kind of documentation and it's documentation may
+serve as an example to keep in mind when considering approaches to tackle
+this problem. [Re-frame's documentation](https://github.com/Day8/re-frame)
+is also an interesting case as it is spread between it's README and other
+files in the `doc/` directory of the project.
+
+Generally I believe a good README on Github is desirable regardless of
+more, likely more extensive, documentation elsewhere. Integration of
+such README should be taken into consideration as well.
+
+For a first version Markdown-only would be an acceptable limitation
+but extending to support other markup languages such as Asciidoc
+should be possible.
+
+#### Grimoire
+
+Since we use Grimoire as a storage system it would be nice to
+integrate this non-API information into the Grimoire store as well.
+A few challenges arise:
+
+- Grimoire's support for articles and similar kinds of documentation is unfinished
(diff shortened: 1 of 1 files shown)


danielcompton:
This seems pretty good, a few thoughts:

* I wonder whether the attribute maps should be more like `{:cljdoc/type :markdown :cljdoc/markdown "# Some markdown"}`
* I think it could be good to see a fully worked example of how these docbundles would work on several different projects that have documentation, to see if it is flexible enough for real world use.
* Where/how will this metadata be created? Is this going to be written by library authors or generated after the fact somewhere else?

martinklepsch:
Hey @danielcompton, thanks for reading! :100:

> I wonder whether the attribute maps should be more like `{:cljdoc/type :markdown :cljdoc/markdown "# Some markdown"}`

I considered this and didn't see a lot of value. I might be missing something though? I kind of went with the most basic/simple thing I could come up with but happy to change.

> I think it could be good to see a fully worked example of how these docbundles would work on several different projects that have documentation, to see if it is flexible enough for real world use.

I kind of put that off because it's tedious 😅  You are absolutely right though and writing out those structures should also help during implementation/testing so I feel like we'll need to do that anyways. 

> Where/how will this metadata be created? Is this going to be written by library authors or generated after the fact somewhere else?

I assume by *metadata* you refer to is the data containing the documentation, i.e. the [first code block in the ADR](https://github.com/martinklepsch/cljdoc/blob/adr/doc-bundles/doc/adr/0007-introduce-docbundles-for-non-api-docs.md#decision). The idea is that library authors provide the hierarchy (if any) and this hierarchy, as well as files in the repository, is used to generate the *metadata*. The metadata is self-contained and has no dependencies to the repo, local files or similar (it might depend on remote files but this is still undecided I'd say).

I hope this makes sense. I'll try to write out the structure for yada and maybe another project soon. (If you have any particular projects you think would be interesting to see, I welcome suggestions.)

danielcompton:
> I considered this and didn't see a lot of value. I might be missing something though? I kind of went with the most basic/simple thing I could come up with but happy to change.

Having a more regular map structure with known key types and locations makes it a bit easier to make other tooling that can consume the information. It'll probably also be easier to spec if you go that route.

Everything else looks good though.

martinklepsch:
I think I've been trying to address to many things at once in this ADR and so I edited it to only cover the `library <> cljdoc` interface for specifying a table of contents: https://github.com/martinklepsch/cljdoc/pull/9.

Will close this. Any decision on Grimoire storage etc. will be made at a later point in time.

martinklepsch:
> Having a more regular map structure with known key types and locations makes it a bit easier to make other tooling that can consume the information. It'll probably also be easier to spec if you go that route.

@danielcompton I see what you mean. I was planning on using spec + conformers to figure out what "type" something is but I see your point.

PR #9, opened 2018-01-23
ADR-0007: Introduce DocTrees to Describe Non-API Docs

martinklepsch:
This is an alternative to the ADR proposed in #5. 

In contrast to the decisions made in #5 this ADR focuses on the interface to library authors may use to specify a hierarchy of documentation and does not concern itself with any specifics on how that information is used or stored.

For the near future I'm planning to just save the resulting information in the Grimoire Version meta directly. This should be good enough for first prototypes and leave the door open for improvements coming out of Grimoire.

Link to [Rendered Version of the ADR](https://github.com/martinklepsch/cljdoc/blob/adr/doc-trees/doc/adr/0007-introduce-doctrees-to-describe-non-api-docs.md).

---

**EDIT** FWIW here's some code that can process the DocTree for Yada as outlined in the ADR.
```clojure
(ns cljdoc.doc-trees
  (:require [clojure.java.io :as io]))

(defn read-file [f]
  (-> f slurp (subs 0 20)))

(declare process-toc)

(defn process-toc-entry [base-dir [title attrs & children]]
  (cond-> {:title title}

    (and (:file attrs) (.endsWith (:file attrs) ".adoc"))
    (assoc-in [:attrs :cljdoc/asciidoc] (read-file (io/file base-dir (:file attrs))))

    (and (:file attrs) (or (.endsWith (:file attrs) ".md")
                           (.endsWith (:file attrs) ".markdown")))
    (assoc-in [:attrs :cljdoc/markdown] (read-file (io/file base-dir (:file attrs))))

    (seq children)
    (assoc :children (process-toc base-dir children))))

(defn process-toc [base-dir toc]
  (mapv #(process-toc-entry base-dir %) toc))

(comment
  (clojure.pprint/pprint
   (process-toc (:dir yada) (:toc yada))))

```


doc/adr/0007-introduce-doctrees-to-describe-non-api-docs.md
```diff
@@ -0,0 +1,158 @@
...
+```
+
+## Consequences
+
+#### Monolithic Documentation
+
+Projects like
+[Fulcro](https://github.com/fulcrologic/fulcro/blob/develop/DevelopersGuide.adoc)
+have one big file which contains all documentation. It would therefore
+not be possible for Fulcro to describe a DocTree with items pointing
```
danielcompton: Could Fulcro describe the doctree with hash links to section ids? That's a little bit web centric, but could be an option.

martinklepsch: I considered that and it would be possible. It would would add quite a bit of complexity though as we'd need to "chunk" documents, and therefore have a deeper understanding of the various document types we want to support (Markdown, Asciidoc, ...). Also in contrast to files uniqueness is not guaranteed.

Definitely willing to do this kind of thing if that's whats needed but for v1 I'll see what people say about splitting their docs into multiple files 🙂

PR #11, opened 2018-01-30
UI: Scroll indicator & auto-scroll for def blocks in the sidebar

roman01la:


martinklepsch:
Hey Roman, this looks great, so smooth! 😸 I have a few ideas for improvement but I'm also happy to merge as-is and maybe get back to those later. So here are some thoughts:

- You introduced some classes which don't have any CSS meaning but are only relevant for JS. I've sometimes seen people prefix those with `.js-` or similar. What do you think about that approach?
- I really like that the `.sidebar-scroll-view` is also scrolled along with the main content section. 🙌 
  When scrolling the main container all the way up after scrolling down the sidebar could also be scrolled to the top, currently it is only scrolled so that the topmost `def-item` comes into view.
- What do you think about moving the indicator to the edge of the sidebar like [this](https://user-images.githubusercontent.com/97496/35612922-aa30d2de-066a-11e8-8ae6-a82182a65754.png) or [this](https://user-images.githubusercontent.com/97496/35612927-ad2b1094-066a-11e8-806e-728bc1a043fa.png)?
- I think it would be nice to separate the "figuring out where there scroll indicator should start/end" from the actual rendering. This would probably make it easier to explore ideas like the one above. 
- Sometimes with long def-names I noticed that the scroll indicator affects text-flow. In combination with the separation suggestion above we could also turn it into one absolutely positioned element which should not affect text-flow.

Let me know what you think about those suggestions and thanks so much already for contributing 🤗

roman01la:
> You introduced some classes which don't have any CSS meaning but are only relevant for JS. I've sometimes seen people prefix those with .js- or similar. What do you think about that approach?

I'm fine with the prefix. Let's do this.

> When scrolling the main container all the way up after scrolling down the sidebar could also be scrolled to the top, currently it is only scrolled so that the topmost def-item comes into view.

Sure, will do. Just did it the easy way :)

> What do you think about moving the indicator to the edge of the sidebar like this or this?

Moving the indicator to the edge of the window makes it less noticeable and when it's on the right edge of the sidebar it seems to be too distant from defblock links which makes it harder to align which items are matching the indicator, IMO. I'll try both and see what looks/feels best.

> I think it would be nice to separate the "figuring out where there scroll indicator should start/end" from the actual rendering. This would probably make it easier to explore ideas like the one above.

Do you mean the actual code? Probably, I'll see what can be done.

> Sometimes with long def-names I noticed that the scroll indicator affects text-flow. In combination with the separation suggestion above we could also turn it into one absolutely positioned element which should not affect text-flow.

Yeah, that's a bit of a problem. I'll move it into an absolute positioned element as you've suggested. It would be a bit more work, but should be much more robust.

martinklepsch:
Merged this manually, live now: https://cljdoc.xyz/d/bidi/bidi/2.1.2/api/bidi.bidi

Issue #18, opened 2018-04-06
Highlighting of code blocks from Markdown/Asciidoc


martinklepsch:
Title says it all :wink: Any code that is rendered as part of a Markdown/Asciidoc document should be rendered similar to how it is done on GitHub.

eerohele:
I'd like to take a crack at this.

A minimal patch that does the job (for Markdown, at least — didn't test AsciiDoc yet) is:

```diff
--- a/src/cljdoc/renderers/html.clj
+++ b/src/cljdoc/renderers/html.clj
@@ -17,10 +17,17 @@
                  [:title (:title opts)]
                  [:meta {:charset "utf-8"}]
                  [:link {:rel "stylesheet" :href "https://unpkg.com/tachyons@4.9.0/css/tachyons.min.css"}]
-                 (hiccup.page/include-css "/cljdoc.css")]
+                 (hiccup.page/include-css
+                   "https://cdn.jsdelivr.net/gh/highlightjs/cdn-release@9.12.0/build/styles/github-gist.min.css"
+                   "/cljdoc.css")]
                 [:div.sans-serif
                  contents]
-                (hiccup.page/include-js "/cljdoc.js")]))
+                (hiccup.page/include-js
+                  "https://cdn.jsdelivr.net/gh/highlightjs/cdn-release@9.12.0/build/highlight.min.js"
+                  "https://cdn.jsdelivr.net/gh/highlightjs/cdn-release@9.12.0/build/languages/clojure.min.js"
+                  "https://cdn.jsdelivr.net/gh/highlightjs/cdn-release@9.12.0/build/languages/clojure-repl.min.js"
+                  "/cljdoc.js")
+                [:script "hljs.initHighlightingOnLoad();"]]))

 (defn sidebar-title [title]
   [:h4.ttu.f7.fw5.tracked.gray title])
```

Some notes / questions, though:

- I used jsDelivr instead of unpkg here because we can't use the Node.js version of Highlight.js.
- Another option would be to do the syntax highlighting server-side with something like [Clygments](https://github.com/bfontaine/clygments). That would probably speed up page loading slightly, but it's a bit more effort. I think a custom `NodeRenderer` would do the job, though, and I can certainly look into it. Thoughts?
- I selected the GitHub Gist Highlight.js style for now. I think it fits the current style OK and I didn't want to spend too much time on the CSS yet since I don't know what the plans are style-wise.
[...]

PR #26, opened 2018-05-18
Add syntax highlighting via Highlight.js #18

eerohele:
Highlight.js is the quickest and easiest way to add syntax highlighting
support that works for Markdown, AsciiDoc, and potential other
documentation formats.

I investigated server-side syntax highlighting via Clygments
(https://github.com/bfontaine/clygments), but it currently doesn't
support guessing the language of the code block.

Furthermore, the AsciiDocJ extension API proved to be quite fiddly.
Using it to reliably convert all AsciiDoc codeblocks would probably be a
somewhat larger effort.

Highlight.js is able to guess the code block language, but the downside
is that the default distribution provides support for 23 of the most
common programming languages. Adding support for every new language
comes at the cost of increased network payload for users.

### DIFF
--- a/resources/public/cljdoc.css
+++ b/resources/public/cljdoc.css
@@ -71,3 +71,9 @@
   bottom: 0;
   left: -2rem;
 }
+
+/** Namespace Page Code Blocks -------------------------------------- **/
+
+.ns-page .def-block .hljs {
+  padding: inherit;
+}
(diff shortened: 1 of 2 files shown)


bfontaine:
> I investigated server-side syntax highlighting via Clygments (https://github.com/bfontaine/clygments), but it currently doesn't support guessing the language of the code block.

FYI I just released a new version of Clygments that supports this feature. However, in my experience Pygment’s guesses aren’t always accurate, especially on small code snippets.

PR #45, opened 2018-05-25
Use the commit SHA if the release isn't tagged. Fixes #43.

### LINKED ISSUE #43: Source links should fall back to pom.xml sha if no git tag is present
Source links for projects that don't tag their releases on GitHub are currently broken. We could either fall back to the revision specified in the SHA or we could even use that as a default since tags are mutable. (I do like having the version in the URL though 🙂)

(Source links are the links to the source on GitHub for each def in a namespace.)

karls:
Hey @martinklepsch!

Took a quick stab at the source links issue. LMK what you think.


martinklepsch:
Thanks a bunch! 👏

PR #55, opened 2018-05-30
Use zprint to format arglists to fit within available space.

kkinnear:
As written, this uses a hard-coded 70 column width, since that seems to be about the available space.  The default of 80 was definitely too big, things were still cut off.  If there is some configuration for width of the html output, I'd be happy to integrate with that.  I tested this on generating API documentation for zprint itself (which has some pretty big arglists) and they came out fine.

Oh, one small thing.  The latest repo I cloned failed when I ran it because there was no `data` directory, but once I created it by hand, it worked fine.  That seems to be a new thing, as the previous version I tried this on didn't have that problem.  No big deal, the error is quite clear.

### DIFF
--- a/build.boot
+++ b/build.boot
@@ -42,7 +42,8 @@
     [ragtime "0.7.2"]
 
     [expound "0.6.0"]
-    [metosin/bat-test "0.4.0" :scope "test"]])
+    [metosin/bat-test "0.4.0" :scope "test"]
+    [zprint "0.4.9"]])
 
 (boot.core/set-env! :source-paths #{"src" "test"}
                     :resource-paths #{"resources"}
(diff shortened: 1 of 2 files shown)


martinklepsch:
Looks amazing @kkinnear! Look at it! https://cljdoc.xyz/d/preflex/preflex/0.4.0/api/preflex.resilient 🎉 

Made a small tweak here https://github.com/martinklepsch/cljdoc/commit/2a503479353e03a8b77d74aa02ccfba35cde788f to ensure protocol methods are also nicely formatted 👍 

Thanks a bunch, this makes a big difference!

martinklepsch:
Oh, and while I got so excited about nicely formatted argument lists, I almost forgot the data directory thing. Pushed a fix here: 1057499. Thanks for bringing this up 🙏 

If you have any further ideas for improving the first-run experience, please let me know! 🙂

PR #61, opened 2018-06-03
@alex-dixon/@timothypratley: adds maven central as a repository

timothypratley:
Clojure is not hosted on Clojars.
Including Maven Central in the repositories list allows Clojure to be
documented. Note that clojure.parallel fails, but the docs are still
generated for other namespaces. clojure.parallel is deprecated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,6 @@ ops/.terraform
 ops/terraform.tfstate*
 ops/image/image-id
 ops/prod-backup/
+.idea
+*.iml
+project.clj
--- a/build.boot
+++ b/build.boot
@@ -61,7 +61,7 @@
          '[cljdoc.config :as cfg]
          '[cljdoc.analysis.task :as ana]
          '[cljdoc.util]
-         '[cljdoc.util.clojars :as clojars]
+         '[cljdoc.util.repositories :as repositories]
          '[cljdoc.util.boot])
 
 (spec/check-asserts true)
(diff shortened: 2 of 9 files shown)


alex-dixon:
https://github.com/timothypratley/cljdoc/pull/1
- Added tests to cover some of the changes
- Renamed cljdoc.util.clojars to cljdoc.util.repositories

> Did I understand correctly that analyzing Clojure itself will just work with these changes?

Yep! We saw the generated documentation for all clojure namespaces minus `clojure.parallel`. Codox fails on that because it requires an external dependency on the classpath. The documentation for everything else is generated just fine though. 

> I guess it might still be good to hardcode the exclusion of clojure.parallel

Adding `clojure.parallel` to `hardcoded-projects-config.edn` gave us the same result as we have now -- still an error with `clojure.parallel`. Maybe Codox evaluates all namespaces before ignoring them?

martinklepsch:
Awesome you added some tests! I considered suggesting this but after my own track record of adding tests to this project I didn’t dare to ask[1] 🙈

Will get this merged some time tomorrow,  excited about getting Clojure itself on cljdoc! 

Not sure how much you looked at the rest of the code base but if you have any suggestions for improvements feel free to suggest them here or on Slack. Working on a code base mostly alone sometimes makes you oversee certain deficiencies 😌

[1] im mostly done with a refactor to switch from yada to pedestal and am planning to add some end to end tests as part of that, fingers crossed 🤞

PR #65, opened 2018-06-07
Badges 🛡

martinklepsch:
This PR adds two things:

- an endpoint that redirects to the latest released version's documentation

      https://cljdoc.xyz/jump/release/reagent
- an endpoint to get a badge showing that version

      https://cljdoc.xyz/badge/reagent

### DIFF
--- a/doc/userguide/faq.md
+++ b/doc/userguide/faq.md
@@ -13,3 +13,17 @@ Source Control Management (SCM) info is information about where the source of yo
 
 1. It's nice because it shows up on Clojars and people will find your project more easily.
 2. cljdoc uses your Git repo to gather additional information, without the SCM info we can only show API documentation and cannot provide a backlink to your Git repo.
+
+# Can we have badges?
+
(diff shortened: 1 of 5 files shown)


martinklepsch:
Deployed, here's a nice little reagent badge:

[](https://cljdoc.xyz/jump/release/reagent)

martinklepsch:
Sometimes shields.io responds a little to slow for GitHub and we get a `504 Gateway Timeout`. Allowing the badge to be cached for some time helped a bit but really we should set a short `max-age` and set an `ETag` header. If the version changes we could check that with the `If-None-Match` header. Unfortunately I haven't been able to understand when exactly GitHub sends the `If-None-Match` header so this will have to wait a bit.

- Some [good docs on HTTP caching](https://devcenter.heroku.com/articles/increasing-application-performance-with-http-cache-headers)
- The [badge rendering template](https://github.com/badges/shields/blob/707ada37189582db837f7a1e563aa01afc919070/templates/flat-template.svg) should we need it later.

PR #93, opened 2018-08-08
Switched to sort by line versus name

### LINKED ISSUE #89: Sort vars by occurrence (line number)
Currently vars are sorted by name, this is the most basic way of sorting stuff but we also have line numbers and the order functions are defined can often provide some useful context.

Let's sort stuff by line number instead of name.

**Question** Should this also apply to ["index listings"](https://cljdoc.xyz/d/compojure/compojure/1.6.1/api/compojure)?

jdhorwitz:
Switched from sorting by name to sorting by line.

If this is not correct, or there are better different things I am missing please let me know.

Fixes #89

### DIFF
--- a/src/cljdoc/bundle.clj
+++ b/src/cljdoc/bundle.clj
@@ -33,7 +33,7 @@
   [all-defs ns]
   (->> all-defs
        (filter #(= ns (:namespace %)))
-       (group-by :name)
+       (group-by :line)
        (vals)
        (map platf/unify-defs)
        (sort-by platf-name)))
--- a/src/cljdoc/platforms.clj
+++ b/src/cljdoc/platforms.clj
@@ -56,7 +56,7 @@
 (defn unify-defs [platforms]
   (let [clean-members (fn clean-members [members]
                         (->> members
-                             (sort-by :line)
+                             (sort-by :name)
                              (map #(dissoc % :file :line))))]
     ;; todo assert names are equal
     (doseq [p platforms]
--- a/src/cljdoc/platforms.clj
+++ b/src/cljdoc/platforms.clj
@@ -56,7 +56,7 @@
 (defn unify-defs [platforms]
   (let [clean-members (fn clean-members [members]
                         (->> members
-                             (sort-by :name)
+                             (sort-by :line)
                              (map #(dissoc % :file :line))))]
     ;; todo assert names are equal
     (doseq [p platforms]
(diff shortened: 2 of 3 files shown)


martinklepsch:
Hey @jdhorwitz, the change looks good, thanks for your contribution :clap: 👍 

I wonder if we should maybe add a little note "Order as defined in namespace." or similar at the top of the list so that it's clear and people know what to expect. 

I've also been thinking that ordering by line shows "utility" functions first which may be less interesting than "API" functions that make use of those utility functions later on. I think we'll have to try and see how it goes 🙂

jdhorwitz:
I like the note of "Order as defined in namespace".  I'll add that in to this PR.

jdhorwitz:
@martinklepsch Where were you thinking would be a good place to put that note?

martinklepsch:
I think below or above the platform support note would be fine👍

jdhorwitz:
Thanks @martinklepsch Let me know how it looks!

martinklepsch:
Hey @jdhorwitz, sorry for taking very long on this. There are a couple of issues here that I only realized after merging and deploying your changes 😅 

1. The place where you changed the sorting isn't actually the place where all vars of a namespace get sorted. Instead that bit of code is used to "merge" information about vars for different platforms like Clojure / ClojureScript. I added a docstring to that function now but also read the namespace docstring to understand what's going on there. FWIW I'm not sure why anything is being sorted there at all. 
2. The right place to change sorting is here: 
      https://github.com/cljdoc/cljdoc/blob/65f4ae93f36d82fcb4f24ab596295854bed6baaa/src/cljdoc/bundle.clj#L32-L39

3. Last but not least a bit of a bummer. Because we deal with vars from multiple platforms we might have the same var with varying source `:line` fields. For instance the var could be defined in a `.clj` and a `.cljs` file and the order in those files is completely independent. All ideas I had to work around this had one or more flaws:
   - Consistently getting the `:line` field for platform `cli` and using that to sort will fail if there is a var that does not exist on `cljs`.
   - Consistently getting the `:line` field from `cli` and falling back to `cljs` if a var only exists there could work but also feels a little bit random. 

I'm not sure what we should do here.

bbatsov:
I wonder if this shouldn't be made configurable, though. One can argue that both sorting schemes have their merits.

martinklepsch:
Hey Josh,

sorry this PR has been lingering for so long. The original issue I outlined above still exists but I think we could do the following:

- Add a link to the top of the list of vars. This link should show either `Order by definition` or `Order alphabetically` depending on the currently selected mode.
- Persist the selection in a session cookie or similar.
- When rendering use the value of the cookie to determine sort order.

Not sure if a cookie is the right thing to use here, if anyone has other suggestions, please share.

martinklepsch:
Closing this for now :)

PR #106, opened 2018-08-15
Fix wikilink markup reference

bbatsov:
Otherwise this was just resolved as an empty link.


martinklepsch:
Him, I don't think that should be necessary. Can you provide the example where you ended up with an empty link?

bbatsov:
I tried to fix

image

Seems I failed.

bbatsov:
You're welcome! I only wish my adoc mojo was stronger. 😆

PR #117, opened 2018-09-22
added toggle raw

Tavistock:
Sometimes the renderer mangles some quasi md in docstrings. I started to think about extending the md parser but i think it's an intractable problem. so instead add a option to just render the raw docstring. Motivating example in the images from `re-frame.std-interceptors`'s docstring:

screen shot 2018-09-22 at 3 20 25 pm

screen shot 2018-09-22 at 3 20 58 pm

### DIFF
--- a/resources/public/cljdoc.css
+++ b/resources/public/cljdoc.css
@@ -118,16 +118,3 @@
   bottom: 0;
   left: -2rem;
 }
-
-.raw {
-    font-family: "SFMono-Regular", Consolas, "Liberation Mono", Menlo, Courier, monospace;
-    padding: 1rem;
-    background-color: rgb(246, 248, 250);
-    border-radius: 3px;
-    font-size: 0.875rem;
-    overflow-x: scroll;
-}
-
-.hide {
-    display: none;
-}
--- a/resources/public/cljdoc.css
+++ b/resources/public/cljdoc.css
@@ -118,3 +118,16 @@
   bottom: 0;
   left: -2rem;
 }
+
+.raw {
+    font-family: "SFMono-Regular", Consolas, "Liberation Mono", Menlo, Courier, monospace;
+    padding: 1rem;
+    background-color: rgb(246, 248, 250);
+    border-radius: 3px;
+    font-size: 0.875rem;
+    overflow-x: scroll;
+}
+
+.hide {
+    display: none;
+}
--- a/resources/public/cljdoc.js
+++ b/resources/public/cljdoc.js
@@ -3,7 +3,8 @@
 var NSPage = document.querySelector(".ns-page")
 
 if (NSPage) {
-  initSrollIndicator()
+    initSrollIndicator()
+    initToggleRaw()
 }
 
 function initSrollIndicator() {
@@ -40,6 +41,28 @@ function initSrollIndicator() {
   mainScrollView.addEventListener("scroll", drawScrollIndicator)
 
   drawScrollIndicator()
+}
+
+function initToggleRaw() {
+    var toggles = Array.from(document.querySelectorAll(".toggle-raw"))
+
+    function addToggleHandlers() {
+        toggles.forEach((el, idx) => {
+            el.addEventListener("click", function () {
+                var parent = el.parentElement
+                var markdown = parent.querySelector(".markdown")
+                var raw = parent.querySelector(".raw")
+                if (markdown.classList.contains("hide")) {
+                    markdown.classList.remove("hide")
+                    raw.classList.add("hide")
+                } else {
+                    markdown.classList.add("hide")
+                    raw.classList.remove("hide")
+                }
+            })
+        })
+    }
+    addToggleHandlers()
 }
 
 window.onbeforeunload = function(){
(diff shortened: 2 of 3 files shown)


src/cljdoc/render/api.clj
```diff
@@ -32,12 +32,14 @@
...
-       (rich-text/markdown-to-html
+  [:div
+   [:div.lh-copy.markdown
+    (-> doc-str
+        (rich-text/markdown-to-html
          {:escape-html? true
           :render-wiki-link (comp render-wiki-link parse-wiki-link)})
-       hiccup/raw)])
+        hiccup/raw)]
+   [:pre.lh-copy.raw.hide (hiccup/raw doc-str)]])
```
martinklepsch: `hiccup/raw` will prevent escaping of HTML in the passed string, I think in this case we do want any HTML in the doctoring to be escaped.

resources/public/cljdoc.css
```diff
@@ -118,3 +118,16 @@
...
+    padding: 1rem;
+    background-color: rgb(246, 248, 250);
+    border-radius: 3px;
+    font-size: 0.875rem;
+    overflow-x: scroll;
+}
+
+.hide {
+    display: none;
+}
```
martinklepsch: cljdoc mostly uses Tachyons (see [ADR #0011](https://github.com/cljdoc/cljdoc/blob/master/doc/adr/0011-use-tachyons-as-css-foundation.md)) for styling, basically using utility classes right in the markup. I think in this case we probably want something like the following:

- `.code`([tachyons docs](http://tachyons.io/docs/typography/font-family/)) for monospaced font, which is also whats used for arglists. Although an argument could be made that your stack is nicer since it uses SFMono when available. 
- `.pa3` 1rem padding
- `.br2` for 4px border-radius
- `.f6` for font-size
- `.overflow-x-scroll` for, well... overflow-x: scroll

For the background color you can see if you find something here: https://tachyons.io/docs/themes/skins/

finally you can eliminate the `.hide` class by using the `display` helpers: 

- `.dn` display none
- `.db` display block

resources/public/cljdoc.js
```diff
@@ -42,6 +43,28 @@ function initSrollIndicator() {
...
+                    raw.classList.add("hide")
+                } else {
+                    markdown.classList.add("hide")
+                    raw.classList.remove("hide")
+                }
+            })
+        })
+    }
+    addToggleHandlers()
+}
```
martinklepsch: The file uses 2 spaces for indentation so far which might cause trouble the next time someone auto indents it. Do you think we should switch to 4 spaces or keep 2? I realize 2 is somewhat non-standard in JS 🙂

src/cljdoc/render/api.clj
```diff
@@ -100,7 +102,8 @@
            [:a.link.f7.gray.hover-dark-gray.mr2
             {:href (platf/get-field def :src-uri p)}
             (format "source (%s)" p)])
-         [:a.link.f7.gray.hover-dark-gray {:href (platf/get-field def :src-uri)} "source"]))]))
+         [:a.link.f7.gray.hover-dark-gray.mr2 {:href (platf/get-field def :src-uri)} "source"]))
+     [:a.link.f7.gray.hover-dark-gray.toggle-raw {:href "#"} "toggle raw"]]))
```
martinklepsch: It would be nice if the button would alternate between `raw docstring` / `formatted docstring` depending on the state of the UI.

martinklepsch: In some other places there has been a loose convention of prepending JS-relevant classnames with `js--` so I guess we could also make this one `js--toggle-raw`?

martinklepsch:
I want to make merging of PRs as easy as possible for contributors while still providing feedback that will allow people to contribute more effectively in the future if they choose to. 

With that in mind we can 

1. Merge this PR as is and I'll make the few adjustments I requested in the comments
2. Make changes while this PR is open until it's 'ready'

Let me know what you prefer, I'm happy about this contribution either way 🤗

Tavistock:
option 2 is fine. I'll address all the issues in the comments and push up some new commits.

resources/public/cljdoc.js
Tavistock: Spacemacs defaulted to 4. I have no js style opinions. I suggest we go with the original style of 2.

src/cljdoc/render/api.clj
Tavistock: done and done

resources/public/cljdoc.css
Tavistock: I updated the styling to use tachyons

src/cljdoc/render/api.clj
Tavistock: I removed the call which should escape html

martinklepsch:
@Tavistock merged this now, feel free to open more PRs for the namespace docstring 👍 thanks!

martinklepsch:
I just deployed this, thanks again very much Travis! Here's a quick demo: https://giant.gfycat.com/SpiritedUnhappyLaughingthrush.mp4

PR #125, opened 2018-10-04
Show toggle for `raw docstring` when docstring exists

### LINKED ISSUE #119: Don't show "raw docstring" switch if there is no docstring
e.g. here: https://cljdoc.xyz/d/re-frame/re-frame/0.2.0/api/re-frame.core

saskali:
fixes #119 

**issue to be solved:** 
at the moment, the toggle `raw docstring` is always shown, even when no docstring is available

**solution:**
only show the toggle when `render-doc` returns something else than `nil`

ideas or suggestions for improvements are welcome 😃

### DIFF
--- a/src/cljdoc/render/api.clj
+++ b/src/cljdoc/render/api.clj
@@ -68,7 +68,8 @@
 (defn def-block
   [def render-wiki-link]
   {:pre [(platf/multiplatform? def)]}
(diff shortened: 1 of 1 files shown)


martinklepsch:
Whoop Whoop 🎉 Thanks a lot Saskia! 🙂 

Merged and deployed! 🚀

PR #129, opened 2018-10-04
Mention clj REPL workflow

jorinvo:
Workflow was mention before but this should clear things up for anyone new to this setup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,8 +54,16 @@ that is done you can reload the page and view the documentation.
 
 Logs can be viewed with `tail -f log/cljdoc.log`.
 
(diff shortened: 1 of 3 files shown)


jorinvo:
Another way would be:
```
clj -e "(require '[cljdoc.server.system]) (in-ns 'cljdoc.server.system)" -r
```
Not sure if that is confusing.

martinklepsch:
🚀 Thanks a lot! 

If you have any further ideas for improving the Getting-Started-Experience please share. It's good to have some fresh eyes every now and then spotting weird things that I have grown accustomed to by now 😄

Issue #128, opened 2018-10-04
Better copy on 


martinklepsch:
When no docs for for an artefact have been built the artefact overview page that usually lists versions doesn't list anything. This is confusing and there should be a few lines of copy that indicate that this page only lists versions that documentation has been built for. 

<img width="831" alt="screen shot 2018-10-04 at 20 32 21" src="https://user-images.githubusercontent.com/97496/46495031-a05ab380-c814-11e8-9c21-45f7c2087473.png">

This may also apply to group ID pages like this one: https://cljdoc.xyz/d/org.clojure

via #124

jorinvo:
I can look into this if you don't mind.

jorinvo:
Guess can close this now :)

martinklepsch:
Good idea :)

martinklepsch:
Good idea :)

Issue #85, opened 2018-07-23
Implement proper search for cljdoc database to replace use of Clojars API


martinklepsch:
The search on the cljdoc.xyz homepage currently only searches Clojars using Clojars' own API. This may result in surprising search results when search for artefacts like Clojure itself, clojure.spec or similar.

---

#### Option 2: Use some search API for Maven Central and filter the result

---

There are pros and cons to each option. May expand on this later.

Related to some issues outlined in #83

holyjak:
Isn't nr. 1 a quick fix and thus worth doing before eventually implementing nr 2? 

Also, since there already is documentation for (some of?) these libraries in cljdoc (e.g. https://cljdoc.org/d/org.clojure/data.json/), cannot we query cljdoc's own metadata (instead of talking to maven central)? Any pointers to the relevant code? Thanks!

martinklepsch:
@holyjak yes, doing **#1** first would definitely make sense, especially since Clojure artefacts on Maven Central are fairly well known and mostly under `org.clojure/`. 

#### As for how things work currently:

- cljdoc does not have any built in search capabilities. The search on the front page and the search that Dash is using is both powered by Clojars, coming down to essentially this: https://github.com/cljdoc/cljdoc/blob/7eb4fbf9696acaffceb714f3593c9cf8ff97e7ef/js/search.js#L23-L28
- This is problematic because it 
  1. doesn't include stuff from Maven Central and
  2. will often show unexpected results before the thing you're actually looking for, see
    
       ```
       curl 'https://clojars.org/search?q=re-frame&format=json' | jq
       ```

#### How to improve this

(This is just an idea, totally open to other approaches.)

- Set up Lucene to index all projects on Clojars as well as the ones from the manually created list (e.g. for Maven Central)
- Use Lucene for the search on the homepage & work with Dash maintainer to use that as well
- Currently cljdoc's `versions` table **does not contain all artifacts** on Clojars but only the ones that documentation has been built for. So there might be some things to be figured out as to what data sources to use to feed into the index.
- I wrote some code that has been parked in #233 that we could use as input to sort by downloads/popularity (arguably a stretch goal)

If you're interested in helping with this I'd be happy to pair and/or provide more direction as needed.

holyjak:
### 1. Search engine

[...]

PR #131, opened 2018-10-05
Help on empty pages

jorinvo:
This should solve #128.

I did a couple of changes which I think help users when ending up on those empty pages:

- Help user when package has no built versions

  Ask user which version they are looking for and redirect them to the
  appropriate page.

  First I thought about simply explaining which URL the user has to go to
  but this seems friendlier.  
  My second idea was to do the redirect in JS on the client directly, but
  I decided that the extract work is worth it to not depend on JS.

  redirect-to-version is similar to jump-to-project. Maybe we can find
  some way to combine them?

  Also I reused the validation of request-build. Not sure if that's the
 right way.

- Hide "Other artifacts under" if group is empty

- Render "No known artifacts under the group" 

- Add top-bar with home link to empty pages

This is how the group, package and version page would look like:

http://localhost:8000/d/org.clojure
group

http://localhost:8000/d/org.clojure/tools.logging
package

http://localhost:8000/d/org.clojure/tools.logging/1.2.3
version

Please let me know what you think and what I could improve!
Also, please have a look at the code. I am still learning about the Clojure ways of doing things :)

### DIFF
--- a/src/cljdoc/render/layout.clj
+++ b/src/cljdoc/render/layout.clj
@@ -84,6 +84,14 @@
     (into [:div.absolute.top-0.bottom-0.left-0.right-0.overflow-y-scroll.ph4-ns.ph2.main-scroll-view]
           content)])
 
+(defn top-bar-generic []
+  [:nav.pv2.ph3.pv3-ns.ph4-ns.bb.b--black-10.flex.items-center
+   [:a {:href "/"}
+    [:span.link.dib.v-mid.mr3.pv1.ph2.ba.hover-blue.br1.ttu.fw5.f7.silver.tracked "cljdoc Beta"]]
+   [:a.silver.link.hover-blue.ttu.fw5.f7.tracked.pv1
+    {:href (util/github-url :issues)}
+    "Have Feedback?"]])
+
 (defn top-bar [cache-id scm-url]
   [:nav.pv2.ph3.pv3-ns.ph4-ns.bb.b--black-10.flex.items-center
    [:a.dib.v-mid.link.dim.black.b.f6.mr3 {:href (routes/url-for :artifact/version :path-params cache-id)}
--- a/src/cljdoc/renderers/html.clj
+++ b/src/cljdoc/renderers/html.clj
@@ -34,24 +34,26 @@
   (let [group-id (:group-id route-params)
         artifacts (:artifacts cache-contents)
         big-btn-link :a.db.link.blue.ph3.pv2.bg-lightest-blue.hover-dark-blue.br2]
-    (->> [:div.pa4-ns.pa2
-          [:h1 group-id]
-          (if (empty? artifacts)
-            [:span.db "No known artifacts under the group " group-id]
-            [:div
-             [:span.db "Known artifacts and versions under the group " group-id]
-             (for [a artifacts]
-               [:div
-                [:h3 (format "%s/%s" group-id a)]
-                [:ol.list.pl0.pv3
-                 (for [version (->> (:versions cache-contents)
-                                    (filter #(= (:artifact-id %) a))
-                                    (sort-by :version)
-                                    (reverse))]
-                   [:li.dib.mr3.mb3
-                    [big-btn-link
-                     {:href (routes/url-for :artifact/version :path-params (assoc version :group-id group-id))}
-                     (:version version)]])]])])]
+    (->> [:div
+          (layout/top-bar-generic)
+          [:div.pa4-ns.pa2
+           [:h1 group-id]
+           (if (empty? artifacts)
+             [:span.db "No known artifacts under the group " group-id]
+             [:div
+              [:span.db "Known artifacts and versions under the group " group-id]
+              (for [a artifacts]
+                [:div
+                 [:h3 (format "%s/%s" group-id a)]
+                 [:ol.list.pl0.pv3
+                  (for [version (->> (:versions cache-contents)
+                                     (filter #(= (:artifact-id %) a))
+                                     (sort-by :version)
+                                     (reverse))]
+                    [:li.dib.mr3.mb3
+                     [big-btn-link
+                      {:href (routes/url-for :artifact/version :path-params (assoc version :group-id group-id))}
+                      (:version version)]])]])])]]
          (layout/page {:title (str group-id " — cljdoc")
                        :description (format "All artifacts under the group-id %s for which there is documenation on cljdoc"
                                             (:group-id cache-id))}))))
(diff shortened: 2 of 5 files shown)


src/cljdoc/render/layout.clj
```diff
@@ -84,27 +84,35 @@
...
-     [:input.pa2.mr2.br2.ba.outline-0.blue {:type "hidden" :id "version" :name "version" :value (:version cache-id)}]
-     [:input.f7.white.hover-near-white.outline-0.bn.bg-white {:type "submit" :value "rebuild"}]]
-    (if scm-url
-      [:a.link.dim.gray.f6.tr
-       {:href scm-url}
-       [:img.v-mid.mr2 {:src "https://icon.now.sh/github"}]
-       [:span.dib (util/gh-coordinate scm-url)]]
-      [:a.f6.link.blue {:href (util/github-url :userguide/scm-faq)} "SCM info missing"])]])
+(defn top-bar
+  ([]
```
jorinvo: I added this function overload. Not sure if there is a more idiomatic way.

src/cljdoc/server/routes.clj
```diff
@@ -39,6 +39,7 @@
 
 (defn utility-routes []
   #{["/jump/release/*project" :get nop :route-name :jump-to-project]
+    ["/redirect-to-version" :post nop :route-name :redirect-to-version]
```
jorinvo: Really not sure about the naming of the route. And should it be placed under `/jump`?

src/cljdoc/render/layout.clj
martinklepsch: I think an alternative approach could have been to create a separate function `top-bar-generic` (since the difference is that it's not specific to a particular project). 

This would also keep the diff smaller making it potentially easier to follow for reviewers.

src/cljdoc/renderers/html.clj
```diff
@@ -32,22 +32,28 @@
...
-                [:li.dib.mr3.mb3
-                 [big-btn-link
-                  {:href (routes/url-for :artifact/version :path-params (assoc version :group-id group-id))}
-                  (:version version)]])]])]
+    (->> [:div
+          (layout/top-bar)
+          [:div.pa4-ns.pa2
+           [:h1 group-id]
+           (if (empty? artifacts)
+             [:span.db "No known artifacts under the group " group-id]
```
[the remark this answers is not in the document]
jorinvo: sounds good!

src/cljdoc/render/layout.clj
jorinvo: Sure, I can do that! Should I share the markup or is the duplication alright?

jorinvo:
Totally understand your point here. Having a link to the current version would be way more helpful. I am just not sure where to get the current version from. I also know that it will only work for Maven packages once #85 is done. Do you have any points to where I can find the current version?

Also, do you prefer if I keep the commit history for these attempts here or should I amend/cherry-pick for only the final version of the PR?

martinklepsch:
> Do you have any points to where I can find the current version?

Sorry I should have mentioned that 🙈 😄 There is a "pseudo version" `CURRENT` that can be used like this:

```
https://cljdoc.xyz/d/org.clojure/tools.logging/CURRENT 
```

Also for context: the search on the front-page only searches Clojars because Clojars exposes a JSON-based search API. The CURRENT mechanism does not rely on this search API and instead talks to the Maven repository directly (Clojars and Maven Central can both be considered Maven repositories). See [`cljdoc.util.repositories`](https://github.com/cljdoc/cljdoc/blob/master/src/cljdoc/util/repositories.clj) if you want to better understand what's going on.

> Also, do you prefer if I keep the commit history for these attempts here or should I amend/cherry-pick for only the final version of the PR?

No strong opinions here, feel free to do what's easiest for you. I might squash/rearrange commits while merging if it makes sense (e.g. when a commit has mostly been reverted).

src/cljdoc/renderers/html.clj
```diff
@@ -56,28 +62,43 @@
...
-            [:div
-             [:h3 "Other artifacts under the " (:group-id cache-id) " group"]
-             [:ol.list.pl0.pv3
-              (for [a (sort (:artifacts cache-contents))]
-                [:li.dib.mr3.mb3
-                 [big-btn-link
-                  {:href (routes/url-for :artifact/index :path-params (assoc cache-id :artifact-id a))}
-                  a]])]])]
+        artifacts (:artifacts cache-contents)
+        versions (->> (:versions cache-contents)
```
jorinvo: Extract `artifacts` and `versions` in let

src/cljdoc/renderers/html.clj
```diff
@@ -56,28 +62,43 @@
...
+                      (filter #(= (:artifact-id %) (:artifact-id route-params)))
+                      (sort-by :version v/version-compare)
+                      (reverse))
+        btn-link :a.dib.bg-blue.white.ph3.pv2.br1.no-underline.f5.fw5
```
jorinvo: Button styles, please feel free to make it prettier!

src/cljdoc/renderers/html.clj
```diff
@@ -56,28 +62,43 @@
...
+        big-btn-link :a.db.link.blue.ph3.pv2.bg-lightest-blue.hover-dark-blue.br2
+        ]
+    (->> [:div
```
jorinvo: Wrapped in a div to include `(layout/top-bar-generic)`.

src/cljdoc/renderers/html.clj
```diff
@@ -56,28 +62,43 @@
...
+          (layout/top-bar-generic)
+          [:div.pa4-ns.pa2
+           [:h1 (util/clojars-id artifact-entity)]
+           (if (empty? versions)
```
jorinvo: Render link on empty versions

src/cljdoc/renderers/html.clj
```diff
@@ -56,28 +62,43 @@
...
+              ]
+             [:div
+              [:span.db "Known versions on cljdoc:"]
+              [:ol.list.pl0.pv3
+               (for [v versions]
+                 [:li.dib.mr3.mb3
+                  [big-btn-link
+                   {:href (routes/url-for :artifact/version :path-params (merge cache-id v))}
+                   (:version v)]])]])
+           (when-not (or (empty? artifacts) (= #{artifact-id} (set artifacts)))
```
jorinvo: Hide when artifacts empty

jorinvo:
I refactored my code to use `top-bar-generic`, adapted the message on the group page and replaced the form + API endpoint with the way simpler solution of linking to the current version:


screen shot 2018-10-06 at 14 21 00

I tried commenting the parts I changed since it is rather difficult to see in old this markup which parts actually changed. Please let me know, if there is anything else to improve!

martinklepsch:
Thanks for this @jorinvo, and congrats on your <strike>first</strike>, err, second! contribution to cljdoc 🎉 🍻 

I decided to squash the commits because the first few commits have been revised quite a bit.

I also used the pretty awesome [Autochrome](https://github.com/ladderlife/autochrome) to get a more readable diff for this change and it worked very nicely. You can see [the diff here](https://cdn.rawgit.com/martinklepsch/59642160207e4cb1657fda3a68a0c793/raw/d2ab166bc0282662f8d31eb79cdfa4de25d99bea/cljdoc-131-diff.html).

PR #132, opened 2018-10-05
Change switcher shortcut k -> 'cmd/ctrl + k'

avichalp:
Change switcher shortcut from `k` to `cmd` + `k` on Mac and `ctrl` + `k` on windows.
Use keydown events to correctly handle key combinations.

### DIFF
--- a/resources/public/switcher.js
+++ b/resources/public/switcher.js
@@ -66,9 +66,18 @@
   }
 
   class Switcher extends Component {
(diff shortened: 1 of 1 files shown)


martinklepsch:
Thanks a lot & congrats on your first contribution @avichalp! 🎉 

I think your decision on using `keydown` makes perfect sense 👍

PR #136, opened 2018-10-08
traverse all available docstring blocks in toggleRaw

### LINKED ISSUE #127: raw docstring toggle does not work if there are different docstrings in clj/s
In some cases we have different docstrings per platform and the docstring-toggle does not traverse the entire list of `.raw` blocks. Example: https://cljdoc.xyz/d/re-frame/re-frame/0.10.5/api/re-frame.interop#reagent-id

I would also suggest to rename `.raw` to `.js--raw-docstring` for better grep-ability and clearer indication that it's JS-related.

Here is the code that needs to traverse all available `.raw` blocks:

https://github.com/cljdoc/cljdoc/blob/1f2b7d084cb211250a9e2edf57ec73b98f30bf0f/resources/public/cljdoc.js#L53-L70

greg-kargin:
screen shot 2018-10-08 at 23 57 39

Fixes #127

### DIFF
--- a/resources/public/cljdoc.js
+++ b/resources/public/cljdoc.js
@@ -48,23 +48,26 @@ function initSrollIndicator() {
 }
 
 function initToggleRaw() {
-  var toggles = Array.from(document.querySelectorAll(".js--toggle-raw"))
+  let toggles = Array.from(document.querySelectorAll(".js--toggle-raw"));
 
   function addToggleHandlers() {
-    toggles.forEach((el, idx) => {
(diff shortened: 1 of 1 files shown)


martinklepsch:
Looks great Greg! Thanks & congrats to your first contribution to cljdoc 🎉 

If you send me your address via email (see GH profile) I'd be happy to send you a few cljdoc stickers :)

martinklepsch:
Deployed now as well. (You may need to clear your cache.) 🍻

PR #141, opened 2018-10-11
Set autofocus property on search field

### LINKED ISSUE #133: Focus on search bar when page loads
Most common use case for a user on the home page is to search for the library. It might be a good idea to focus on the search bar when page is loaded. 

[docs.rs](https://docs.rs) also does that.

mk:
Closes #133.


martinklepsch:
Sweet, thanks a bunch! 🍻 

I also have some stickers for you if you want them 😄

PR #152, opened 2018-10-17
JS Modules + Packaging

### LINKED ISSUE #122: Clean up the JavaScript setup
cljdoc uses [a few JS files](https://github.com/cljdoc/cljdoc/tree/master/resources/public) as well as some dependencies (react + a fuzzy search lib). Currently this isn't packaged in any way and files are simply loaded. While this works reasonably well it makes reuse harder and forces us to load files in a specific order so their dependencies are met. 

To fix this we should consider adding a packager of some kind. Webpack, Parcel or [Fastpack](http://fastpack.io) seem like reasonable choices.

- Commands to bundle assets once and in "watch mode" should be documented in the [Running cljdoc locally guide](https://github.com/cljdoc/cljdoc/blob/master/doc/running-cljdoc-locally.md).
- The `run-cljdoc-api.sh` script will also require some adjustments so that this build step is ran before the server is started. 
- In order to avoid confusion about compiled vs. handwritten resources all compiled files should be stored in a separate directory in the root of the project `resources-compiled`.

martinklepsch:
Fixes #122 

⚠️ This PR required some changes to the deployment tooling and so I'm now also pushing a lot of changes related to that.

This PR refactors all JS into ES6 modules and provides some basic Parcel bundling to be used for production bundling and development builds.

```
./script/parcel watch
./script/parcel build
```

The compiled JS is placed on the classpath via `resources-compiled` a new top level directory intended to contain files that are generated as part of the build process.

Some things still to be done: 

- [x] Investigate how to get source maps working. Parcel requests `/index.map` but should request `/js-compiled/index.map`. I did some quick googling but couldn't find anything helpful.
- [x] Update the documentation for running cljdoc locally to include this new step.
- [x] Integrate the parcel building into the production environment. 
   - [x] make sure parcel is installed in prod env
   - [x] run it as part of `run-cljdoc-api.sh`

Eventually we should also rethink the packaging of cljdoc in general. As more build steps are introduced it might make sense to not run all of them when starting up the application and instead move them to some phase where a Jar or similar gets build.


script/parcel
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -eou pipefail
+in="js/index.js"
+out="resources-compiled/public/js/"
+
+echo "parcel $1 $in -d $out"
+parcel $1 "$in" -d "$out"
```
rakyi: One alternative common in JS world would be to define commands such `watch` and `build` in `"scripts"` property in package.json and invoke them by e.g. `npm run build`, see https://docs.npmjs.com/cli/run-script

PR #154, opened 2018-10-17
transform #"..." to #regex"..." in anlayze-result

### LINKED ISSUE #143: Cljdoc build failes when a regex is used in the arglist (as default value in :or)
I tried to generate the documentation for https://github.com/metasoarous/semantic-csv.
The [build](https://circleci.com/gh/martinklepsch/cljdoc-builder/5452#tests/containers/0) failed with 'No dispatch macro for: "'.

The reason is, that in [this function](https://github.com/metasoarous/semantic-csv/blob/4b2447175d3de501155254038133559f56de14cf/src/semantic_csv/transducers.cljc#L232) an regex is used as a default value. 

This regex ends up in the arg-list in the file produced by the analyser, that is read in again in [the runner](https://github.com/cljdoc/cljdoc/blob/67fbc69c0f0d78f32c7914dd2761307f46d0f591/modules/analysis-runner/src/cljdoc/analysis/runner.clj#L95) and causes this error (as a regex is not valid edn).

IamDrowsy:
transform #"..." to #regex"..." in the analyze-result.edn and read them in again with a custom reader tag
Fixes #143

### DIFF
--- a/modules/shared-utils/src/cljdoc/util.clj
+++ b/modules/shared-utils/src/cljdoc/util.clj
@@ -6,7 +6,8 @@
   what's on the classpath."
   (:require [clojure.edn]
             [clojure.java.io :as io]
-            [clojure.string :as string])
+            [clojure.string :as string]
+            [clojure.edn :as edn])
   (:import (java.nio.file Files)))
 
 (def hardcoded-config
@@ -35,6 +36,28 @@
   [project version]
   {:pre [(some? project) (string? version)]}
   (str "cljdoc-edn/" (group-id project) "/" (artifact-id project) "/" version "/cljdoc.edn"))
+
+(defn- wrap-regex-tag [potential-regex]
+  (if (instance? java.util.regex.Pattern potential-regex)
+    (tagged-literal 'regex (str potential-regex))
+    potential-regex))
+
+(defn- wrap-all-regex-tags [structure]
+  (clojure.walk/postwalk wrap-regex-tag structure))
+
+(defn- unwrap-regex-tag [regex-tagged-string]
+  (re-pattern regex-tagged-string))
+
+(defn serialize-cljdoc-edn [analyze-result]
+  ;; the analyzed structure can contain regex #"..." (e.g. in :arglists)
+  ;; and they can't be read in again using edn/read-string
+  ;; so there are changed to #regex"..." and read in with a custom reader
+  (-> analyze-result
+      (wrap-all-regex-tags)
+      (pr-str)))
+
+(defn read-cljdoc-edn [file]
+  (edn/read-string {:readers {'regex unwrap-regex-tag}} (slurp file)))
 
 (defn git-dir [project version]
   (str "git-repos/" (group-id project) "/" (artifact-id project) "/" version "/"))
--- a/modules/shared-utils/src/cljdoc/util.clj
+++ b/modules/shared-utils/src/cljdoc/util.clj
@@ -4,17 +4,17 @@
   These are available in the analysis environment and thus should work
   without any additional dependencies or further assumptions about
   what's on the classpath."
-  (:require [clojure.edn]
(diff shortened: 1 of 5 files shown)


modules/shared-utils/src/cljdoc/util.clj
```diff
@@ -36,6 +37,28 @@
...
+(defn serialize-cljdoc-edn [analyze-result]
+  ;; the analyzed structure can contain regex #"..." (e.g. in :arglists)
+  ;; and they can't be read in again using edn/read-string
+  ;; so there are changed to #regex"..." and read in with a custom reader
+  (-> analyze-result
+      (wrap-all-regex-tags)
+      (pr-str)))
+
+(defn read-cljdoc-edn [file]
+  (edn/read-string {:readers {'regex unwrap-regex-tag}} (slurp file)))
```
martinklepsch: The redirection here makes the code slightly harder to read IMHO. What do you think about?
```suggestion
  (edn/read-string {:readers {'regex re-pattern}} (slurp file)))
```

modules/shared-utils/src/cljdoc/util.clj
```diff
@@ -36,6 +37,28 @@
...
+
+(defn- unwrap-regex-tag [regex-tagged-string]
+  (re-pattern regex-tagged-string))
+
+(defn serialize-cljdoc-edn [analyze-result]
+  ;; the analyzed structure can contain regex #"..." (e.g. in :arglists)
+  ;; and they can't be read in again using edn/read-string
+  ;; so there are changed to #regex"..." and read in with a custom reader
+  (-> analyze-result
+      (wrap-all-regex-tags)
```
martinklepsch: IMHO having the code spread across multiple functions makes it slightly harder to read than if it were all in one function. How would you feel about this? 

```clojure
(->> analyze-result
     (clojure.walk/postwalk #(if (instance? java.util.regex.Pattern %)
                               (tagged-literal 'regex (str %))
                               %)))
     (pr-str))
```

modules/shared-utils/src/cljdoc/util.clj
```diff
@@ -6,7 +6,8 @@
   what's on the classpath."
   (:require [clojure.edn]
             [clojure.java.io :as io]
-            [clojure.string :as string])
+            [clojure.string :as string]
+            [clojure.edn :as edn])
```
martinklepsch: `clojure.edn` has already been required in it's fully qualified form above.
`clojure.walk` is missing a require. 🙂

modules/shared-utils/src/cljdoc/util.clj
IamDrowsy: IMHO having this indirection only makes sense when the `wrap-regex-tag` is it's own function, as it makes it clear that they are opposites.

modules/shared-utils/src/cljdoc/util.clj
IamDrowsy: Usally when threading data, I tend to make all the steps into separate functions as it can be read more fluently top to bottom without figuring out what every single call exactly does (athough most of the time finding good names is the hard part 🙂). Having only one step, I'm totally fine with not having a separate function 🙂 and will change the PR following your suggestion.

modules/shared-utils/src/cljdoc/util.clj
IamDrowsy: whoops. missed that.

martinklepsch:
Thanks! Merged this manually in https://github.com/cljdoc/cljdoc/commit/353409d9c6a1918eff0743b82bbd1cc907e37e31. 🎉

martinklepsch:
@IamDrowsy I added you to the `cljdoc` org and you should now have push access to cljdoc. 👍 

If feel like it you could also add some tests for the new read/serialize functions 😄

PR #158, opened 2018-10-18
Change primary domain from cljdoc.xyz to cljdoc.org

### LINKED ISSUE #135: consider a TLD other than xyz
Often .xyz domains are associated with malware site. I work at Cisco Systems and they are automatically blocked :) Not sure if other networks would have similar policies.

danielcompton:
I'm not sure if I got everything here, but it seems like a reasonable start. Rather than trying to do the redirects in Clojure, I added a new server config for the xyz and 301 redirected to the new domain.

api.cljdoc.org also needs to be handled, but that doesn't seem to be currently working (SSL error), so it might not need to be done at the same time.

Fixes #135

### DIFF
--- a/ops/cljdoc.tf
+++ b/ops/cljdoc.tf
@@ -8,7 +8,6 @@ provider "aws" {
   secret_key = "${var.aws_secret_key}"
 }
 
-
 provider "digitalocean" {
   token = "${var.do_token}"
 }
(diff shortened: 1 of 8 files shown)


martinklepsch:
Thanks a lot for this Daniel! I've been somewhat hesitant to approach the change but your effort here is a good encouragement 😄 

`api.cljdoc.xyz` is effectively deprecated. Feel free to remove it or I will do so before merging.

martinklepsch:
Went all in on this today, migration is done changes are in #177 — closing this.

PR #165, opened 2018-10-18
Only run the deploy on master branch

danielcompton:
This prevents deploys from running in pull requests (which fail because env vars aren't exported) and on cljdoc branches (which would presumably succeed).


martinklepsch:
Thanks Daniel! 🙌 

I originally thought deploying branches is nice but it only works for cljdoc branches and I'm not sure how to configure that. Also it's not that important 😄

Issue #171, opened 2018-10-20
cljdoc not running local on windows


IamDrowsy:
Currently cljdoc cannot run on windows because:
1. There is no `clj` for windows
2. https://github.com/weavejester/ragtime/issues/103

1. can be worked around by starting cljdoc via repl, but you will still need a way to consume the `deps.edn` on windows (the only way I'm currently aware of is the cursive 1.8.0-eap2 build).

2. can be worked around by adding [this code](https://github.com/weavejester/ragtime/issues/103#issuecomment-253727091) to the start up as long as the issue is not solved.

martinklepsch:
We should probably take over [that ragtime PR] if the original author doesn't get back to it. (Which seems likely-ish with a year passed.)

As for using `clj` on windows / shelling out... would be interesting to talk to Sean Corfield and other Windows users, I feel like there must be a way to get this to work? 😄 

The jar URI thing should be easy to fix as well... I guess?

martinklepsch:
Windows WSL was mentioned as being useful to run tools-deps on Windows: https://docs.microsoft.com/en-us/windows/wsl/install-win10 

Proper Windows support would still be great but just wanted to note this for future Windows contributors 🙂

IamDrowsy:
Small status update:
* cljdoc runs fine on WSL, so contributing on Windows 10 is definitly possible
* [the ragtime issue](https://github.com/weavejester/ragtime/issues/103) is solved and should be fixed in `0.8.0`
* with the new ragtime version, the server now starts from the repl, but there are small encoding issues so you should start the repl with `-Dfile.encoding=UTF-8` (these are not new, but I missed them initially).
* still no `clj/clojure` in windows, there are alternatives like [clj-windows](https://github.com/frericksm/clj-windows), but I haven't looked into them.

As WSL is working, I probably won't dig further into the remaining issue until an official `clj/clojure` is available.

martinklepsch:
Thanks for keeping track of this!

Please don’t hesitate to make any changes to master as you see fit (updating ragtime, documenting that encoding property, ...).

Thanks 🙏

JulienRouse:
Small update on this:

Sean Corfield added recently a wiki page [for Windows user](https://github.com/clojure/tools.deps.alpha/wiki/clj-on-Windows)  of [tools.deps.alpha](https://github.com/clojure/tools.deps.alpha). That is still in alpha but hopefully it will be workable!

IamDrowsy:
[...]

PR #172, opened 2018-10-20
on windows, string of URI is not a valid file,...

IamDrowsy:
but io/file of URI should be valid on all plattforms.
small part of #171


IamDrowsy:
Whoops, I need to look at this again later.

Issue #137, opened 2018-10-09
Navigate search results using up/down arrow keys


avichalp:
I think it will be better UX to let users navigate search results with arrow keys (up and down).
Also using `return` key on a result to open its docs.

martinklepsch:
Good catch @avichalp! For the quick switcher that is available via <kbd>CMD</kbd> + <kbd>k</kbd> I did implement the up/down arrow thing. 

https://github.com/cljdoc/cljdoc/blob/63a99b91de40b51ce235c474f3e73b62bc302135/resources/public/switcher.js#L91-L95

I think in general there is a lot of code that could be shared between those two pieces (search & quick switcher) but I didn't want to abstract too much before that emerged more clearly.

greg-kargin:
I tried to tackle this issue, but for some reason search-widget and quick-switcher don't appear on page when i run local instance of clj-doc.

martinklepsch:
Oh and thanks for taking a look at it 🙌

martinklepsch:
@greg-kargin back at a computer now, here are the docs to setup the JS stuff: https://github.com/cljdoc/cljdoc/blob/master/doc/running-cljdoc-locally.md

It would be kind of nice if in development we would render some small error message in place of the JS components if the JS file could not be found. Not sure if it's worth the effort — What do you think?

greg-kargin:
@martinklepsch, thanks for such detailed response! I agree that showing error-plaque in place of missing components is a great idea. I'll try to fix both of this issues 👍

PR #174, opened 2018-10-20
notify when javascript files were not compiled

greg-kargin:
screen shot 2018-10-20 at 23 55 01

feature discussed in https://github.com/cljdoc/cljdoc/issues/137


martinklepsch:
@greg-kargin Thanks for this! 🙌 

I'm wondering if this will show while the JS is being loaded? Might be confusing for people if an error like that flashes on the screen and then disappears. Have you noticed this while testing?

greg-kargin:
You are right. That's not that bad when ran locally on my machine, blinks for couple ms in 1/10 reloads, but I can see cases when it might cause users frustration 😐. I see two options in fixing this:

- make warning less noticeable
- add warning only in development mode

First approach is straight-forward to implement. For second one I need some insights on how to pass a development flag to a hiccup component.

greg-kargin:
You are right. That's not that bad when ran locally on my machine, blinks for couple ms in 1/10 reloads, but I can see cases when it might cause users frustration 😐. I see two options in fixing this:

- make warning less noticeable
- add warning only in development mode

First approach is straight-forward to implement. For second one I need some insights on how to pass a development flag to a hiccup component.

greg-kargin:
You are right. That's not that bad when ran locally on my machine, blinks for couple ms in 1/10 reloads, but I can see cases when it might cause users frustration 😐. I see two options in fixing this:

- make warning less noticeable
- add warning only in development mode

First approach is straight-forward to implement. For second one I need some insights on how to pass a development flag to a hiccup component.

greg-kargin:
You are right. That's not that bad when ran locally on my machine, blinks for couple ms in 1/10 reloads, but I can see cases when it might cause users frustration 😐. I see two options in fixing this:

- make warning less noticeable
- add warning only in development mode

First approach is straight-forward to implement. For second one I need some insights on how to pass a development flag to a hiccup component.

greg-kargin:
You are right. That's not that bad when ran locally on my machine, blinks for couple ms in 1/10 reloads, but I can see cases when it might cause users frustration 😐. I see two options in fixing this:

- make warning less noticeable
- add warning only in development mode

First approach is straight-forward to implement. For second one I need some insights on how to pass a development flag to a hiccup component.

martinklepsch:
There’s a profile defined in `cljdoc.config` probably the easiest way is to require that and only show the warning if the profile isn’t “prod”.

I usually don’t like loading config into random namespaces like this and instead prefer to pass it as explicit arguments but in this case it seems fine since it’s just a development thing...

rakyi:
Not that I'm against this addition, but isn't documenting the build steps in combination with the error in JS console enough? There are multiple places where JS components are/will be used and adding static HTML to warn that JS is not running is pretty non-standard and might get pretty unwieldy, pretty soon.

greg-kargin:
Current readme.md is a bit misleading, as it suggests to run script and says nothing about javascript compilation. Also there is no link to docs in contributing section, so I didn't even know that they exist before @martinklepsch sent me a link. 
I think that adding a link to documentation and mentioning javascript in contributing section might be a good idea and that there is nothing particularly wrong in adding this warning message to static html in dev-mode, as it might give a clue to such inattentive developers as I am 😃 .

martinklepsch:
Good points about the confusing docs folks, I updated the README to point to the canonical way to get up and running: https://github.com/cljdoc/cljdoc#contributing — let me know if this could use further improvements.

I also had another idea for this problem, which boils down to this:

1. in development add another (hidden) div to the body containing an error
1. have an inline script that checks if `/js/index.js` can be loaded
1. if it can't make the div visible

I have a basic version here: https://github.com/cljdoc/cljdoc/compare/no-js-warning — feel free to use some of that code if you think it makes sense. Something that's still missing from this branch is the condition that this is only added in development.

greg-kargin:
@martinklepsch, sure no problem:)

PR #176, opened 2018-10-21
only include deps of scope 'provided`, 'system' and 'test'

### LINKED ISSUE #147: Only explicitly pass "provided" dependencies, not regular ones
In order to have "add-on" namespaces with optional dependencies not break the build cljdoc includes all dependencies listed in a project's POM. By doing so we include dependencies with the scope "provided" that would otherwise not be included by the dependency resolution. 

This is the code defining what dependencies will specified explicitly:

https://github.com/cljdoc/cljdoc/blob/a04892814a08d1ceaa55b76dcaa2b821ec060fbd/modules/shared-utils/src/cljdoc/util/deps.clj#L43-L60

The issue here is now that we not only explicitly depend on dependencies with scope provided but also dependencies that would be included by regular resolution. In the process of doing so some information like the `<type>` property of a dependency can get lost.

[clj-boost's POM](http://repo.clojars.org/clj-boost/clj-boost/0.0.2/clj-boost-0.0.2.pom) is a good example for this since it contains a dependency with a `<type>` field:

```xml
<dependency>
  <groupId>ml.dmlc</groupId>
  <artifactId>xgboost4j-example</artifactId>
  <version>0.80</version>
  <type>pom</type>
</dependency>
```

Related to #54 to some extent.

A **possible solution** would be to ensure that `extra-deps` does not return dependencies that have a scope other than `provided`. 

To test this locally this snippet might be handy:

```sh
# at the root of the cljdoc git repo:
./script/analyze.sh clj-boost 0.0.2 ~/.m2/repository/clj-boost/clj-boost/0.0.2/clj-boost-0.0.2.jar ~/.m2/repository/clj-boost/clj-boost/0.0.2/clj-boost-0.0.2.pom
```

IamDrowsy:
also print analyzer result before reading the cljdoc edn, because this fails when the analyzer failed.

Fixes #147

### DIFF
--- a/modules/shared-utils/src/cljdoc/util/deps.clj
+++ b/modules/shared-utils/src/cljdoc/util/deps.clj
@@ -47,7 +47,7 @@
   file but this situation being an edge case this is a sufficient fix for now."
   [pom]
   (->> (pom/dependencies (pom/parse (slurp pom)))
-       (keep (fn [{:keys [group-id artifact-id version]}]
+       (keep (fn [{:keys [group-id artifact-id version scope]}]
                (when-not (or (.startsWith artifact-id "boot-")
                              ;; Ensure that tools.reader version is used as specified by CLJS
                              (and (= group-id "org.clojure")
@@ -55,7 +55,9 @@
                              ;; The version can be nil when pom's utilize dependencyManagement - this unsurprisingly breaks tools.deps
                              ;; Remains to be seen if this causes any issues
                              ;; http://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Management
-                             (nil? version))
+                             (nil? version)
+                             ;; compile/runtime scopes will be included by the normal dependency resolution.
+                             (not (#{"provided" "system" "test"} scope)))
                  [(symbol group-id artifact-id) {:mvn/version version}])))
        (into {})))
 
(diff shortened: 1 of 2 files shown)


modules/shared-utils/src/cljdoc/util/deps.clj
```diff
@@ -47,15 +47,17 @@
...
                              ;; Ensure that tools.reader version is used as specified by CLJS
                              (and (= group-id "org.clojure")
                                   (= artifact-id "tools.reader"))
                              ;; The version can be nil when pom's utilize dependencyManagement - this unsurprisingly breaks tools.deps
                              ;; Remains to be seen if this causes any issues
                              ;; http://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Management
-                             (nil? version))
+                             (nil? version)
+                             ;; compile/runtime scopes will be included by the normal dependency resolution.
+                             (not (#{"provided" "system" "test"} scope)))
```
martinklepsch: Looks great @IamDrowsy! 👍 I find the keep condition `when-not` `or` `not` slightly confusing, maybe it would make sense to refactor those into separate `filter` calls for readability?

IamDrowsy: Just the new line? e.g. adding something like `(filter #(#{"provided" "system" "test"} (:scope %)))` before the `keep` or do you want the refactor the whole `when-not`?

martinklepsch: Whatever you think is best, we can also keep it this way and defer to fix later. 😄

IamDrowsy:
I think you are right that `when-not` `or` `not` is confusing, so at I moved the scope filter to an extra `filter`. For the rest I could not think of any really better version. So I think we should it just leave it as it is.

PR #179, opened 2018-10-22
Bold artifact coordinates for request-build-page

danielcompton:
This makes it clearer which parts cljdoc is using for searching and can help show people why they might have gone wrong. Adding the "v" in front of the version by itself looks slightly confusing. It's not obvious that it was added in the formatting, but wasn't included in the search.

screenshot of google chrome 23-10-18 9-38-47 am


martinklepsch:
Thanks for this Daniel! 

You should also send me your address via email or something so I can send you some cljdoc stickers 

🙂

PR #182, opened 2018-10-22
Use the CURRENT url path for homepage projects

danielcompton:
This will stop it from being out of date as new releases are made.

We could also do a DB query here, but it's probably ideal to not need a DB query to load the homepage.


martinklepsch:
Thank, good catch 🙌

PR #188, opened 2018-10-28
[WIP] Meta concerns ui

saskali:
### What's left to do? 

- implement links for `keyboard shortuts` and `recent improvements` 

#### What keyboard shortcuts are currently available? 

I'm aware of `cmd + k` to jump to most recent docs, are there more? Any ideas for how to display the shortcuts? I was thinking of opening them in a modal, but open for inspiration

#### What has recently changed on cljdoc?

Not sure which place would be best to link here? 

#### Downloading documentation

Should there be some sort of instruction for how to download documentation? Similar question like before if showing a modal, linking to documentation page or something else? 

#### What does this look like on mobile? 

Open question if this ui element should be shown on mobile at all. Here are some screenshots for how it looks like on my phone (iPhone 7, Firefox). 

##### meta info icon
cljdoc-meta-mobile

##### meta info dialog

cljdocmetaexpmobile


avichalp:
Hey @saskali!
Thanks for working on this.

I am trying answers some of the open questions:

> What keyboard shortcuts are currently available?

We currently have only one shortcut at the top level i.e.`CMD+K`.  Opening a modal for displaying shortcuts sounds good to me. 

I also liked the [docs.rs](https://docs.rs/about) approach. It has an about page. May be meta information can go there? like keyboard shortcuts, recent changes/new features, feedback etc.
Also with this approach we don't have to think about do we want to show the help (`?`) symbol on small screens. What do you think?

saskali:
thanks for your feedback, @avichalp! I think an about page is generally a good idea. Looking at what we need for cljdoc at the moment, a modal with meta concerns seems to be a good solution to me because you can stay on the current page when looking for keyboard shortcuts and the other meta fields are links to other pages and don't contain any information themselves. Any thoughts?

martinklepsch:
I’ll give this a more thorough review later today but we could also just have one page with keyboard shortcuts that will open in a new tab. This way we avoid implementing modals for now and can defer this to later if it actually becomes something worth improving. 

For the changelog we dont have anything yet so maybe a reasonable way would be to just comment it out until that changes.

Regarding mobile: My vote goes to not showing for now. Keyboard shortcuts aren’t relevant and screen space is at a premium in general.

martinklepsch:
Oh; and thanks so much for working on this 🙌😊

src/cljdoc/render/layout.clj
```diff
@@ -78,11 +78,50 @@
    {:style {:top (str TOP-BAR-HEIGHT "px") :left "16rem"}} ; CSS HACK
    contents])
 
+(defn meta-info-icon []
+  [:img#meta-icon.ma3
+   {:src "/meta-info-icon.svg"
```
martinklepsch: We've been using icons via https://icon.now.sh in some places, which is pretty nice because there's a lot of stuff and we don't have to maintain/select icons on a one by one basis. There's various help icons but also some other interesting ones that might work in this situation like this [explore one](https://icon.now.sh/explore/48/357edd).

src/cljdoc/render/layout.clj
```diff
@@ -78,11 +78,50 @@
...
+    :alt "meta info icon"
+    :width "60px"
+    :style {:position "fixed"
+            :right 0
+            :bottom 0
+            :background "white"}}])
```
martinklepsch: You can use Tachyons classes for this too: `.fixed.right-0.bottom-0.bg-white`.

For `width` there's various `w-N` classes, probably best to just play around with those and find something that works. (Or maybe just set the size int the icon URL as described on https://icon.now.sh)

src/cljdoc/render/layout.clj
```diff
@@ -78,11 +78,50 @@
...
+
+(defn meta-element [description link]
```
martinklepsch: Since this is only used once I think it could also be moved into `meta-info-dialog`.

js/cljdoc.js
```diff
@@ -126,6 +126,18 @@ function restoreSidebarScrollPos() {
   localStorage.removeItem("sidebarScrollPos");
 }
 
+if (document.querySelector(".main-scroll-view")) {
+  document.getElementById("meta-icon").onclick = function() {
```
martinklepsch: Let's prefix JS-relevant classes with `js--` and avoid too generic names like `close` to make it easier to relate things and avoid conflicts. 

Maybe some thing like:

- `#js--meta-close`
- `#js--meta-dialog`
- `#js--meta-icon`

Also consider exporting this chunk as a function and calling it in `index.js`. Most of the existing JS files just define functions which are then called in `index.js` — this way it's easier to trace the code and see what is happening at a glance.

js/cljdoc.js
```diff
@@ -126,6 +126,18 @@ function restoreSidebarScrollPos() {
...
+    document.getElementById("meta-icon").style.display = "none";
+    document.getElementById("meta-dialog").style.display = "block";
+  };
+
+  document.getElementById("close").onclick = function() {
+    document.getElementById("meta-dialog").style.display = "none";
+    document.getElementById("meta-icon").style.display = "block";
```
martinklepsch: Instead of using `style.display` directly here we could also use tachyons classes which would also allow us to apply media queries — which would be useful if we decide to not show this in smaller screens for now.

For instance using an elements [`classList`](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList) property we could do this:

```js
// to show it:
document.getElementById("meta-icon").classList.replace("dn", "db-ns");
// to hide it: 
document.getElementById("meta-icon").classList.replace("db-ns", "dn");
```

`db-ns` means "**d**isplay: **b**lock if screen **n**ot **s**mall", `dn` means "**d**isplay: **n**one" (and because there's no qualifier like `-ns` it applies to all screen sizes).

src/cljdoc/render/layout.clj
saskali: ahh cool, that's super useful to know! I'll exchange the svg for one of those icons 😃

js/cljdoc.js
saskali: that's super handy, thanks for the suggestion! I'll make that change 👍

saskali:
thanks for the feedback Martin, super valuable and helpful! 
the thing that is left to do now is implementing the page for the keyboard shortcuts. will give that a go within the next days :) 

I noticed the PR for formatting with Prettier, but not entirely sure how it's supposed to work. Should it format JS files with every commit? I did run `npm install` but realised that the tests were failing, so I ran `npm run format` and committed the resulting changes. Let me know if there's a different way of doing this!

martinklepsch:
Hey Saskia! The shortcuts page is looking great! A little empty in a way but nothing wrong with that for now. I rebased your changes against `master` and made some minor adjustments in this branch: [`meta-concerns-ui-tweaks`](https://github.com/cljdoc/cljdoc/compare/master...meta-concerns-ui-tweaks?expand=1). Basically I did the following:

- Increased the clickable area of the buttons in the dialog
- Reduced the icon size to 48px (was 60px)
- Added the `.pointer`  class to the icon to indicate it can be clicked
- Also I squashed some commits that seemed closely related (like fixup ones)

If you don't object any of these changes we can force push the state of the `meta-concerns-ui-tweaks` to this branch and then merge (this way the pull request will be closed properly etc.)

saskali:
looks great, thanks for the quick feedback! no objections from my side, I think this is ready to go in!

martinklepsch:
Yay, it's live :tada: https://cljdoc.org/d/metosin/reitit/0.2.5/doc/introduction

PR #191, opened 2018-10-31
Create a sitemap

### LINKED ISSUE #162: Create a sitemap
I have tried searching for some projects on cljdoc which are not indexed by Google, e.g. "site:cljdoc.xyz cuic" returns nothing, even though it is known. It could be good for cljdoc to build and publish a sitemap of projects so that Google can index them.

dawran6:
Fixes #162 

Here's a summary of what this PR does:

- Add `sitemap` as a dependency
- Add a /sitemap.xml route
- Create ns cljdoc.server.sitemap
- Add a storage API `all-distinct-docs` which queries the versions table for distinct (group_id, artifact_id) pairs
- Resolve urls based on (group_id, artifact_id) pairs and transform them to sitemap.xml

Things yet to be done:
- [x] Change to absolute url
- [x] Fix tests (homepage is 404)
- [x]  Validate the sitemap
- [x] ~~Add :lastmod to sitemap~~
- [x] Change to current versioned URL
- [x] Cache the sitemap (use Cache-Control header, cache in memory, or save on disk, etc)
- [ ] ~~Use the generate-sitemap-and-save*~~

Below is what I got so far. Please feel free to give me your comments. Thanks!

screen shot 2018-10-31 at 11 18 41 am

### DIFF
--- a/src/cljdoc/server/pedestal.clj
+++ b/src/cljdoc/server/pedestal.clj
@@ -11,6 +11,7 @@
             [cljdoc.server.pedestal-util :as pu]
             [cljdoc.server.routes :as routes]
             [cljdoc.server.api :as api]
+            [cljdoc.server.sitemap :as sitemap]
             [cljdoc.storage.api :as storage]
             [cljdoc.util :as util]
             [cljdoc.util.repositories :as repos]
@@ -32,6 +33,11 @@
   (assoc ctx :response {:status 200
                         :body (str body)
                         :headers {"Content-Type" "text/html"}}))
+
+(defn ok-xml! [ctx body]
+  (assoc ctx :response {:status 200
+                        :body (str body)
+                        :headers {"Content-Type" "text/xml"}}))
 
 (def render-interceptor
   {:name  ::render
@@ -301,6 +307,7 @@
   (->> (case route-name
          :home       [{:name ::home :enter #(ok-html! % (render-home/home))}]
          :shortcuts  [{:name ::shortcuts :enter #(ok-html! % (render-meta/shortcuts))}]
+         :sitemap    [{:name ::sitemap :enter #(ok-xml! % (sitemap/sitemap %))}]
          :show-build [pu/coerce-body
                       (pu/negotiate-content #{"text/html" "application/edn" "application/json"})
                       (show-build build-tracker)]
--- a/src/cljdoc/server/pedestal.clj
+++ b/src/cljdoc/server/pedestal.clj
@@ -284,6 +284,24 @@
                                         :body (error/not-found-404)})
               context))})
 
+(defn sitemap-cache
+  "This is a stateful function that caches the sitemap in memory for 60 minutes."
+  [storage state]
+  (let [prev (:last-generated @state)
+        now  (java.util.Date.)]
+    (when (or (not prev)
+              (> (- (.getTime now) (.getTime prev)) 360000))
+      (swap! state assoc :last-generated (java.util.Date.))
+      (swap! state assoc :sitemap (sitemap/build storage)))
+    (:sitemap @state)))
+
+(defn sitemap-interceptor
+  [storage]
+  (let [state (atom {:last-generated nil
+                     :sitemap        nil})]
+    {:name  ::sitemap
+     :enter #(ok-xml! % (sitemap-cache storage state))}))
+
 (defn offline-bundle []
   {:name ::offline-bundle
    :enter (fn offline-bundle [{:keys [cache-bundle] :as ctx}]
@@ -307,7 +325,7 @@
   (->> (case route-name
          :home       [{:name ::home :enter #(ok-html! % (render-home/home))}]
          :shortcuts  [{:name ::shortcuts :enter #(ok-html! % (render-meta/shortcuts))}]
-         :sitemap    [{:name ::sitemap :enter #(ok-xml! % (sitemap/sitemap storage))}]
+         :sitemap    [(sitemap-interceptor storage)]
          :show-build [pu/coerce-body
                       (pu/negotiate-content #{"text/html" "application/edn" "application/json"})
                       (show-build build-tracker)]
--- a/src/cljdoc/server/pedestal.clj
+++ b/src/cljdoc/server/pedestal.clj
@@ -284,23 +284,22 @@
                                         :body (error/not-found-404)})
               context))})
 
-(defn sitemap-cache
-  "This is a stateful function that caches the sitemap in memory for 60 minutes."
-  [storage state]
-  (let [prev (:last-generated @state)
-        now  (java.util.Date.)]
-    (when (or (not prev)
-              (> (- (.getTime now) (.getTime prev)) 360000))
-      (swap! state assoc :last-generated (java.util.Date.))
-      (swap! state assoc :sitemap (sitemap/build storage)))
-    (:sitemap @state)))
+(defn build-sitemap
+  "Build a new sitemap if previous one was built longer than 60 minutes ago."
+  [{:keys [last-generated sitemap] :as state} storage]
+  (let [now (java.util.Date.)]
+    (if (or (not last-generated)
+            (> (- (.getTime now) (.getTime last-generated)) 360000))
+      ;; Return updated state
+      {:last-generated now :sitemap (sitemap/build storage)}
+      ;; Return identical state
+      state)))
 
 (defn sitemap-interceptor
   [storage]
-  (let [state (atom {:last-generated nil
-                     :sitemap        nil})]
+  (let [state (atom {})]
     {:name  ::sitemap
-     :enter #(ok-xml! % (sitemap-cache storage state))}))
+     :enter #(ok-xml! % (:sitemap (swap! state build-sitemap storage)))}))
 
 (defn offline-bundle []
   {:name ::offline-bundle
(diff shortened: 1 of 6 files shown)


danielcompton:
This relates to #162.

src/cljdoc/server/sitemap.clj
```diff
@@ -0,0 +1,30 @@
...
+            [cljdoc.server.routes :as routes]))
+
+
+(defn sitemap [db-spec]
+  (generate-sitemap
+   (for [{:keys [group_id artifact_id] :as doc} (storage/all-distinct-docs db-spec)]
+     {:loc
+      (routes/url-for :artifact/current :path-params {:group-id group_id
+                                                      :artifact-id artifact_id})}))
+  )
```
danielcompton: Can we run the `sitemap.validator` over this at some point, maybe in tests at least, if not also in prod?

src/cljdoc/server/sitemap.clj
```diff
@@ -0,0 +1,30 @@
...
+            [clojure.tools.logging :as log]
+            [cljdoc.config :as cfg]
+            [cljdoc.storage.api :as storage]
+            [cljdoc.server.routes :as routes]))
+
+
+(defn sitemap [db-spec]
+  (generate-sitemap
+   (for [{:keys [group_id artifact_id] :as doc} (storage/all-distinct-docs db-spec)]
+     {:loc
```
danielcompton: Lets put a `:lastmod` in here with the date it was last modified.

src/cljdoc/server/sitemap.clj
```diff
@@ -0,0 +1,30 @@
...
+      (routes/url-for :artifact/current :path-params {:group-id group_id
```
danielcompton: We want this URL to be the current versioned URL, not the root URL which will redirect to the versioned URL.

src/cljdoc/server/pedestal.clj
```diff
@@ -299,6 +305,7 @@
    {:keys [route-name] :as route}]
   (->> (case route-name
          :home       [{:name ::home :enter #(ok-html! % (render-home/home))}]
+         :sitemap    [{:name ::sitemap :enter #(ok-xml! % (sitemap/sitemap storage))}]
```
danielcompton: This is going to be relatively expensive to compute and render, can we think about some form of caching here? At the least, we should set some caching headers, but I think it could also be good to look at caching in-memory the sitemap for some number of minutes, maybe 60? Another idea would be to save it to disk at some location and serve it from there. You could even have nginx serve it up, and have cljdoc update it in a background process.

src/cljdoc/server/sitemap.clj
```diff
@@ -0,0 +1,30 @@
+(ns cljdoc.server.sitemap
+  (:require [sitemap.core :refer [generate-sitemap]]
+            [clojure.tools.logging :as log]
+            [cljdoc.config :as cfg]
+            [cljdoc.storage.api :as storage]
+            [cljdoc.server.routes :as routes]))
+
+
+(defn sitemap [db-spec]
+  (generate-sitemap
```
danielcompton: We may want to use the `generate-sitemap-and-save*` variant at some point. If we start listing sitemaps for individual doc pages then we will quickly blow past the 50k limit in a single sitemap. This probably doesn't have to be here now, I think any kind of sitemap is going to be a massive SEO improvement over having none.

dawran6: Thanks for the comments @danielcompton ! The `generate-sitemap-and-save*` function doesn't exists before version 0.3.0. (https://github.com/hashobject/sitemap/pull/4/files) However the latest on maven central is only 0.2.5. What should I do in this case?

martinklepsch:
Re: https://github.com/cljdoc/cljdoc/pull/191#discussion_r229924048

> What's the purpose of releases table and the versions table? I'd like to know if join the two tables make sense.

The two tables could probably be merged into one. The reason there are two different ones is that SQLite was used to track releases and builds before versions where also stored in SQLite.... and when they were moved into SQLite I didn’t quite realize how close these two are. 

There is a timestamp in the releases table which refers to when it was released to Clojars. We could use that for `lastmod` I guess.

martinklepsch:
@dawran6 this is looking pretty solid already! 

For the caching stuff you could consider creating a stateful interceptor, i.e. one that closes over an atom. That atom could contain a timestamp and the Sitemap only rebuilding the sitemap if the timestamp is older than 60min or similar. 

```clj
(defn sitemap-cache []
  (let [state (atom {})]
    {:enter ,,,}))
```
Saving it to disk in a background job as Daniel suggested is also a reasonable approach :)

dawran6:
Thanks for the comments! I am also uncertain on how to build the absolute url. I am trying to build the function below in the cljdoc.server.routes. However, I'm currently passing in the hardcoded host, port and scheme. What's the better way to get those from the configuration?

```clojure
(def absolute-url-for
  (url-for-routes (routes identity {:host "localhost" :port 8000 :scheme "https"}) :absolute? true))
```

dawran6:
Hi @danielcompton @martinklepsch I made a commit, 3655a3f, to cache the sitemap in memory for 60 minutes. Would you take a look if it looks all right? Thanks!

danielcompton:
I think it would be good to have a way to validate the sitemap with the validators built-in to the sitemap library. Not sure how fast/slow it is, but it could be good to do it when recreating the sitemap after the cache expires. If it fails validation then throw the error to Sentry.

dawran6:
@martinklepsch thanks I think yours look much better. I'll take your advice.
@danielcompton thanks I didn't know what the desired behavior is when the error happens. Shall we returned the previous validated sitemap if validation failed with the current one?

martinklepsch:
> Shall we returned the previous validated sitemap if validation failed with the current one?

I think just throwing an exception and having the server respond with a `500` is probably fine. Crawlers like Google have likely cached the sitemap anyways and we'll fix the issue quickly.

dawran6:
I think the only gap now is that the sitemap data only uses the `versions` table instead of merging it with the `releases` table. And because the `versions` table doesn't have the date, currently it's using the current date.

I'm not sure how to write a SQL query to merge the two tables yet but I'll work on it. Please let me know if you have any thought on this.

```clojure
;;sqlite_impl.clj
(defn all-distinct-docs [db-spec]
  (sql/query db-spec ["select distinct group_id, artifact_id, name from versions"]))
```

martinklepsch:
> I think the only gap now is that the sitemap data only uses the `versions` table instead of merging it with the `releases` table. And because the `versions` table doesn't have the date, currently it's using the current date.

I think it is fine to use the `versions` table for now and not set the `lastmod` field. `lastmod` is useful when the contents are expected to change but most of the content on cljdoc is roughly static. 

Using the `releases` table is problematic since it only contains releases after a certain point in time. So if a user builds docs for a library that has been released in 2017 that will not be in the `releases` table and thus would be missing from the sitemap. Maybe @danielcompton has some ideas how we can get proper release dates for all artefacts on Clojars.

dawran6:
Thanks @martinklepsch ! Just removed the lastmod tag from the sitemap.
I think this PR is ready to be merged if everything seems fine.

danielcompton:
> I think it is fine to use the versions table for now and not set the lastmod field. lastmod is useful when the contents are expected to change but most of the content on cljdoc is roughly static.

I think the benefit of providing `lastmod` is so that the Google crawler knows which pages to prioritise re-crawling. Because we have a large and growing corpus which is mostly static, it'll be useful for crawlers to be able to know which ones to check. The `lastmod` here would just refer to when the content last changed in cljdoc, not necessarily when it was released at Clojars.

martinklepsch:
BTW, `[sitemap "0.3.0"]` is now available.

danielcompton:
> I think however that since we dont use /current (for now) Google will just pick up new pages as they are added to the sitemap.

Ah right, this is the bit I was missing. Yep that's fine then.

src/cljdoc/server/pedestal.clj
```diff
@@ -278,6 +284,23 @@
                                         :body (error/not-found-404)})
               context))})
 
+(defn build-sitemap
+  "Build a new sitemap if previous one was built longer than 60 minutes ago."
+  [{:keys [last-generated sitemap] :as state} storage]
+  (let [now (java.util.Date.)]
+    (if (or (not last-generated)
+            (> (- (.getTime now) (.getTime last-generated)) 360000))
```
danielcompton: I think we need an extra 0 in the last-generated check here? If I'm reading this correctly, 360,000ms is 6 minutes.

```suggestion
            (> (- (.getTime now) (.getTime last-generated)) (* 60 60 1000)))
```

martinklepsch: https://github.com/cljdoc/cljdoc/commit/25c19900b38bdba1d9d364185bace0090ed8cfb6

martinklepsch:
@dawran6 I merged and deployed your changes just now. I also fixed the time issue in the cache conditional that Daniel spotted
(https://github.com/cljdoc/cljdoc/commit/25c19900b38bdba1d9d364185bace0090ed8cfb6). It seems to be [working great](https://cljdoc.org/sitemap.xml) and I also already submitted it to Google. Let's see if this helps with discoverability. 

Thanks a lot 🙌

martinklepsch:
If you want to still upgrade to `[sitemap "0.3.0"]` feel free to do so and either push straight to master or open another PR. Not sure how relevant this is anymore.

src/cljdoc/server/pedestal.clj
dawran6: Whoops right. Missed that one :P Thanks for fixing it!

dawran6:
Awesome! Appreciate all the helps @danielcompton @martinklepsch !

martinklepsch:
My pleasure @dawran6 🙌

PR #196, opened 2018-11-05
navigate search results using keys

### LINKED ISSUE #137: Navigate search results using up/down arrow keys

greg-kargin:
Fixes #137 

Finally had some free time to finish off this issue:).
Now results are also highlighted on hover. Only thing left is to follow list selection with scroll.
Also changed selected element highlight color, as old one was a bit to dim in my opinion.
screen shot 2018-11-05 at 23 52 59

### DIFF
--- a/js/search.js
+++ b/js/search.js
@@ -1,4 +1,5 @@
 import { Component, render, h } from "preact";
+import * as listSelect from "./listselect";
 
 function debounced(delay, fn) {
   let timerId;
@@ -27,6 +28,22 @@ const loadResults = (str, cb) => {
 };
 
 class SearchInput extends Component {
+  onKeyDown(e) {
+    if (e.which === 13) {
+      this.props.onEnter();
+    } else if (e.which === 27) {
+      this.props.unfocus();
+    } else if (e.which === 38) {
+      // arrow up
+      e.preventDefault(); // prevents caret from moving in input field
+      this.props.onArrowUp();
+    } else if (e.which === 40) {
+      // arrow down
+      e.preventDefault();
+      this.props.onArrowDown();
+    }
+  }
+
   render(props) {
     const debouncedLoader = debounced(300, loadResults);
     return h("input", {
@@ -35,7 +52,7 @@ class SearchInput extends Component {
       className: "pa2 w-100 br1 border-box b--blue ba input-reset",
       onFocus: e => props.focus(),
       onBlur: e => setTimeout(_ => props.unfocus(), 200),
-      onKeyDown: e => props.onKeyDown(e),
+      onKeyDown: e => this.onKeyDown(e),
       onInput: e =>
         debouncedLoader(
           cleanSearchStr(e.target.value),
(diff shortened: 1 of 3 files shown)


martinklepsch:
Hey @greg-kargin this is looking good! I was kind of hoping we could unity `search.js` and `switcher.js` since the UI aspects of the two components are pretty much identical but I guess we can also leave this for another day. 

Will merge this later today 👍

greg-kargin:
yep, I think it's fine now. I'm not sure if i will have time to refactor this on the weekends, so I'd better do it in next PR. thanks.

js/search.js
```diff
@@ -45,35 +45,53 @@ class SearchInput extends Component {
...
+      ]
+    )
   ]);
 };
 
 const ResultsView = props => {
   return h(
     "div",
     {
+      id: "results-view",
```
martinklepsch: I'm wondering if it would make sense to use refs here? It seems like the `App` component now has some expectations on the `ResultsView` without that being visible/explicit in any way. This isn't a huge issue for now I guess, just wondering if there's another way?

martinklepsch: Maybe one way to do it would be to introduce a `didUpdate` hook to the `ResultsView`, calling `restrictToViewport` whenever the `selectedIndex` passed via `props` changes.

martinklepsch: No worries in that case, I can probably fix this myself.

martinklepsch:
Hey Greg, not sure if you'll have time but I made some changes on top of yours. If you have the time I'd appreciate a review (writing actual JS is still kinda new to me haha):

- https://github.com/cljdoc/cljdoc/commit/f8fb721e335a9b5318f6156b5bdd2e898ca987a7 moves the scrolling stuff into the `ResultsView`, removing that concern from the overarching `App` component.
- https://github.com/cljdoc/cljdoc/commit/72cff4ebf2db1b91eb036f04c51a414d90529378 updates the `switcher.js` code to use the same `ResultsView` component.

js/search.js
```diff
@@ -124,7 +124,7 @@ class App extends Component {
           resultView: SingleResultView,
           results: state.results,
           selectedIndex: state.selectedIndex,
-          onMouseOver: idx => this.setState({ selectedIndex: idx })
+          onMouseOver: idx => parent.setState({ selectedIndex: idx })
```
greg-kargin: fixed onMouseOver,  `this` refers to function, not component in this case.

js/switcher.js
```diff
@@ -80,6 +80,25 @@ const SwitcherSingleResultView = (r, isSelected, selectResult) => {
...
+        e.preventDefault();
+        this.setState({
+          selectedIndex: Math.min(
+            this.state.selectedIndex + 1,
+            this.state.results.length - 1
+          )
+        });
+      }
+    }
+
```
greg-kargin: moved arrow up/down handling here so it's a little better experience – user can hold arrow on scroll through all list, instead of pushing it down every time

js/switcher.js
```diff
@@ -172,6 +178,12 @@ class Switcher extends Component {
     this.initializeState();
   }
 
+  componentDidUpdate(previousProps, previousState, previousContext) {
+    if (!previousState.show && this.state.show) {
+      this.inputNode.focus();
+    }
+  }
+
```
greg-kargin: Previous solution with `autofocus: true` didn't work in chrome and firefox, this is what I came up with, maybe you know a better way, or it's ok?

js/search.js
greg-kargin: Fixed onMouseOver here, previously `this` referred to function, not component.

js/switcher.js
```diff
@@ -80,6 +80,25 @@ const SwitcherSingleResultView = (r, isSelected, selectResult) => {
 
 class Switcher extends Component {
   handleKeyDown(e) {
+    if (e.target === this.inputNode) {
```
greg-kargin: Moved arrow up/down handling here, so it's a little bit better experience – now selection will move when on hold instead of single-step per keystroke.

js/switcher.js
```diff
@@ -172,6 +178,12 @@ class Switcher extends Component {
     this.initializeState();
   }
 
+  componentDidUpdate(previousProps, previousState, previousContext) {
+    if (!previousState.show && this.state.show) {
+      this.inputNode.focus();
```
greg-kargin: Previous solution with `autofocus: true` didn't work in chrome and firefox. This is what I came up with, what do you think?

greg-kargin:
Hey, @martinklepsch, I reviewed the code everything is great, made some minor additions and described them above. Thanks for help with this PR.
Btw writing JS is kinda new for me too. 😃

martinklepsch:
Thanks for your feedback and additions Greg! 

> Btw writing JS is kinda new for me too. 😃

I guess we're winging it then 😅

greg-kargin:
Thanks for your help Martin! Yep, with JSX it looks better.
I've already sent you my address, but haven't received them yet – russian post is a really slow organisation 🐌

martinklepsch:
Alright, then I guess you gotta be patient some more 😄 Let me know if they don't arrive in the next couple of weeks though!

Issue #107, opened 2018-08-16
Allow passing additional config into CI jobs 


martinklepsch:
Currently all information analysis jobs get is 

- project id
- version
- jar URL
- pom URL

in addition we might want to pass additional configuration like:

- namespaces that should not be included in analysis result (e.g. `cider.nrepl.inlined-deps.*`)
- a hardcoded set of platforms to analyse (some projects contain cljc files but can only be fully loaded in one platform, e.g. reagent only cljs, cider-nrepl only clj)

Unfortunately this is a bit more complex because currently API analysis is done *before* a git repo is even cloned.

PR #197, opened 2018-11-07
Reverse order of analysis

martinklepsch:
The process of building documentation can be thought of as two steps:

1. Analysing a Jar and extracting information about a projects API
2. Analysing a projects Git repository to include a README, provide correct source links etc.

So far the order above has also been the order in which those steps were executed but it turns out that under some circumstances the information from the Git repository may be relevant for the Jar analysis. Primarily users may want to further configure the Jar analysis via the `cljdoc.edn` file in their Git repository (see #107 for more details).

This PR. changes the order of the two steps above, only triggering the CircleCI job after the Git analysis has been completed (successfully or not).

### DIFF
--- a/src/cljdoc/analysis/service.clj
+++ b/src/cljdoc/analysis/service.clj
@@ -8,7 +8,7 @@
 
   Services that implement this interface will receive all relevant information
   via their `trigger-build` method. The expectation then is that the services
-  will eventually call `/api/full-build` with the appropriate params.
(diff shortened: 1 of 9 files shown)


martinklepsch:
I cherry picked a bunch of commits from this PR to master that aren't really related to the analysis reversal. The diff is much smaller now and hopefully should be easier to review as well. 👍

Issue #203, opened 2018-11-12
Don't show SNAPSHOT versions as newer releases


martinklepsch:
Issue with #161 

We should exclude SNAPSHOT versions from the version warning. 

Take a look at the changes done for #161 and this should be an easy fix. 🙂

PR #215, opened 2018-11-17
Fix for #203

elarouss:
excluding snapshot versions from the "new version available" warning


elarouss:
i'm glad it's worked hhhh, of course i want some if it's not burdensome for you

Issue #186, opened 2018-10-25
Search with input form and results page


witek:
The current search on cljdoc.org is not based on web standards. It should be a form with an input field and be backed by a request handler on the server. That way it would be possible to integrate cljdoc searches in browser toolbars.

green-coder:
I found this documentation which seem to cover the intent of having a "custom search engine" in the browser.

https://developer.mozilla.org/en-US/docs/Web/OpenSearch

green-coder:
I am a heavy user of the in-browser search and I am going to take care of this issue.

martinklepsch:
@witek I'll close this issue as I think it mostly addresses your request. Thanks to @green-coder for implementing it 👏

PR #217, opened 2018-11-18
Open search config + search page

green-coder:
References the issue #186 

This PR is not ready to be merged yet. Please provide feedback on the code change.

The search page still need a link for the user to go back to the home page.

### DIFF
--- a/resources/public/opensearch.xml
+++ b/resources/public/opensearch.xml
@@ -1,0 +1,14 @@
+<OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/"
+                       xmlns:moz="http://www.mozilla.org/2006/browser/search/">
+    <ShortName>Cljdoc</ShortName>
+    <Description>Cljdoc is a website building &amp; hosting documentation for Clojure/Script libraries</Description>
+    <InputEncoding>UTF-8</InputEncoding>
+    <Image width="16" height="16" type="image/x-icon">https://cljdoc.org/favicon.ico</Image>
+    <Url type="text/html" template="https://cljdoc.org/search">
+        <Param name="q" value="{searchTerms}"/>
+    </Url>
+    <Url type="application/x-suggestions+json" template="https://cljdoc.org/suggest">
+        <Param name="q" value="{searchTerms}"/>
+    </Url>
+    <moz:SearchForm>https://cljdoc.org/search</moz:SearchForm>
+</OpenSearchDescription>
--- a/src/cljdoc/render/layout.clj
+++ b/src/cljdoc/render/layout.clj
@@ -64,6 +64,10 @@
                    (assert (.startsWith url "/"))
                    [:link {:rel "canonical" :href (str "https://cljdoc.org" url)}]); TODO read domain from config
 
+                 ;; Open Search
+                 [:link {:rel "search" :type "application/opensearchdescription+xml"
+                         :href "/opensearch.xml" :title "Cljdoc"}]
+
                  [:meta {:name "viewport" :content "width=device-width, initial-scale=1"}]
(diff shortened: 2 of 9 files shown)


green-coder:
The PR is now ready to be reviewed and merged if appropriate.

resources/public/opensearch.xml
```diff
@@ -0,0 +1,14 @@
...
+    <ShortName>Cljdoc</ShortName>
+    <Description>Cljdoc is a website building &amp; hosting documentation for Clojure/Script libraries</Description>
+    <InputEncoding>UTF-8</InputEncoding>
+    <Image width="16" height="16" type="image/x-icon">https://cljdoc.org/favicon.ico</Image>
+    <Url type="text/html" template="https://cljdoc.org/search">
+        <Param name="q" value="{searchTerms}"/>
+    </Url>
+    <Url type="application/x-suggestions+json" template="https://cljdoc.org/suggest">
+        <Param name="q" value="{searchTerms}"/>
+    </Url>
```
martinklepsch: would it perhaps make sense to comment that out for now to avoid unnecessary requests?

src/cljdoc/render/layout.clj
```diff
@@ -64,6 +64,10 @@
                    (assert (.startsWith url "/"))
                    [:link {:rel "canonical" :href (str "https://cljdoc.org" url)}]); TODO read domain from config
 
+                 ;; Open Search
+                 [:link {:rel "search" :type "application/opensearchdescription+xml"
+                         :href "/opensearch.xml" :title "Cljdoc"}]
```
martinklepsch: `cljdoc` (all lower case) would be more "on brand" 😄

src/cljdoc/render/search.clj
```diff
@@ -0,0 +1,32 @@
+(ns cljdoc.render.search
+  (:require [cljdoc.render.layout :as layout]
+            [cljdoc.util :as util]))
+
+(defn search-form
+  ([] (search-form ""))
+  ([search-terms]
+   [:div.w-90.mb4
+    [:div#cljdoc-search {:initial-value search-terms}]]))
```
martinklepsch: Maybe `data-initial-value` is more appropriate here, but again, not very important.

src/cljdoc/render/search.clj
```diff
@@ -0,0 +1,32 @@
...
+          (search-form (-> context :request :query-params :q))]]]
+       (layout/page {:title "cljdoc — documentation for Clojure/Script libraries"
+                     :responsive? true})
+       (str)))
+
+;; TODO: complete this function later.
+;; See https://developer.mozilla.org/en-US/docs/Web/OpenSearch for more information.
+(defn suggest-api
+  [context]
+  (assoc context :response {:status 200
```
martinklepsch: Maybe sending a 404 here with no content would be more appropriate until this is implemented?

src/cljdoc/render/search.clj
martinklepsch: https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes

src/cljdoc/render/search.clj
green-coder: Good point. I found the code 501 "not implemented" that fits better, will use it instead.

green-coder:
@martinklepsch I made the code adjustments but I need some access rights to push straight to the master.

resources/public/opensearch.xml
martinklepsch: Thanks for the context, makes sense 👍

martinklepsch:
Thanks for the adjustments!🙌

PR #218, opened 2018-11-19
Improve 404 page

dawran6:
Add explicit 404 message on not-found page:
image

### DIFF
--- a/src/cljdoc/render/error.clj
+++ b/src/cljdoc/render/error.clj
@@ -7,7 +7,8 @@
 (defn not-found-404 []
   (->> [:div.pt4
         [:div.mt5-ns.mt6-l.mw7.center.pa4.pa0-l
-         [:p.f2-ns.f3.mv3.w-90-l.lh-copy "We couldn't find anything here..."]
+         [:h1.f2.f1-ns.b ":("]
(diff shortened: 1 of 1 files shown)


src/cljdoc/render/error.clj
```diff
@@ -7,7 +7,8 @@
 (defn not-found-404 []
   (->> [:div.pt4
         [:div.mt5-ns.mt6-l.mw7.center.pa4.pa0-l
-         [:p.f2-ns.f3.mv3.w-90-l.lh-copy "We couldn't find anything here..."]
+         [:h1.f2.f1-ns.helvetica.b ":("]
```
martinklepsch: Why the `.helvetica` here? cljdoc usually uses `.sans-serif` which defaults to the OS' default sans serif font.

dawran6: I don't have specific reason. Just wanted to play around with different fonts to make it look a bit nicer. Let me put it back.

martinklepsch: Ok cool, was just wondering 🙂

martinklepsch:
Merged, will deploy later today 👍

dawran6:
That's awesome. Thanks!

Issue #223, opened 2018-11-20
Implement the open-search suggestions


green-coder:
When the user set `cljdoc` as the main search engine of the browser (I tested in Firefox only), s/he has the possibility via a checkbox to turn on a "suggestion" feature where cljdoc receives the search terms being typed, and can reply with a list of completions on the terms.

### Usage

1. Setup in the preference tab

![image](https://user-images.githubusercontent.com/598193/48756247-5437d580-ecd3-11e8-8e5f-5eff38489211.png)

2. Type in the address bar

![image](https://user-images.githubusercontent.com/598193/48755941-2d2cd400-ecd2-11e8-803b-9acf3893313e.png)

### Implementation

1. [x] Implementing the suggestion end point correctly.
2. [x] Base the completion upon a list of the most popular libraries of the Clojure(Script) ecosystem, it should be good enough for most of the users.
3. [ ] Improve on the previous part by fetching the data from Clojars once every hour. The completions could be ordered by popularity if we can get this data as well.

holyjak:
Fixed by #316

Issue #208, opened 2018-11-13
Implement some system to cache pom.xml files of projects


martinklepsch:
More interesting data can be integrated into cljdoc such as:

- dependencies and description from POM files
- GitHub Stars and contributor information
- Clojars download stats

This information can be integrated into the cljdoc database but really the canonical datasources for this are very reliable and maybe something that's closer to a caching system would be more appropriate. Especially for data that changes (like download stats or contributors) the time-to-live semantics of a cache could be useful to enforce periodical refreshing.

@avichalp has been working on a cache layer based on `core.cache` and SQLite, initially primarily focused on caching POM files so we can integrate descriptions and dependencies into the cljdoc pages.

**Please chime in if you have thoughts around implementing this.** 

Required for / Related #68

avichalp:
Thanks @martinklepsch for articulating the problem here. Here are some ideas that I have been working on.

#### Why we should have a cache layer?

We want to store some ad-hoc data for eg: Clojars download stats or Github stars etc that doesn't belong in Cljdoc's data model. But storing this data locally saves network calls when we want to read it.

[`clojure.core/memoize`](https://clojuredocs.org/clojure.core/memoize) is great for this job. It caches a function's return values for given arguments. Next time when this function is called with same arguments the value is retrieved from the cache. On top of that [`clojue.core.memoize`](https://github.com/clojure/core.memoize) provides `PluggableCache`  so we can memoize arbitrary functions using out own implementation of a cache. 

In this [PR](https://github.com/cljdoc/cljdoc/pull/239) I have implemented `core.cache`'s CacheProtocol. This cache is persistent. It stores data in a secondary SQLite database. This gives us two benefits:

1. We get to use memoize like api.
2. Data is persisted on disk.

Example on how to memoize a function using this cache:
```clojure
(def db-artifact-uris
    {:key-prefix         "artifact-uris"
     :key-separator      ":"
     :spec {:dbtype      "sqlite"
            :classname   "org.sqlite.JDBC"
            :subprotocol "sqlite"
            :subname     "data/cache.db"}})

(def memoized-artifact-uris
  (memo-sqlite cljdoc.util.repositories/artifact-uris db-artifact-uris)) 

(memoized-artifact-uris 'bidi "2.0.9-SNAPSHOT")
```

Whats left:

- [ ] Write migrations for `Cache` table and `cache.db`.
- [ ] Add `cache-spec` in config .
[...]

Issue #200, opened 2018-11-09
Create list of all Clojure-relevant artifacts on Maven Central


martinklepsch:
There are a bunch of Clojure based artifacts on Maven Central and it would be nice if we could include those in searches as well as check for recent releases that need building on cljdoc.

I think a reasonable way forward could be a mixture of a hardcoded list and something that crawls Maven Central based on group IDs (e.g. https://repo1.maven.org/maven2/org/clojure/).

This might be a somewhat unexciting task and it's low-ish priority for now but I just wanted to open this issue in case anyone feels like working on this or has alternative suggestions about how to approach this.

Required for #85

holyjak:
@martinklepsch do we still need this? Now that we automatically fetch all M.C. artifacts under org.clojure?

martinklepsch:
Good point, I don’t think we need this anymore. 👍

PR #224, opened 2018-11-20
Open search suggestion

green-coder:
Do not merge yet.

Implements the first part of the issue #223. Code review is welcome.


green-coder:
@martinklepsch Some update on this PR:

My next step before I propose to merge this PR is to write a one-time small script that:
- downloads the list of the library names from clojars’ stats endpoints on the last `n` months,
- aggregate them and sum their download counts over the `n` months,
- keep only the `m` most downloaded libraries.

This list will be placed into an `edn` file, slurped into the server's RAM, and used in place of my current library list placeholder. This is intended to be a better temporary solution until we have a proper access to those stats in the DB of the server.

I may also adapt the suggestion's logic to take into account the number of downloads w.r.t. the way we choose the suggestion list.

green-coder:
## WIP
image

Lucene on real-world data (15000+ library names)
The search does not take into account the library's popularity yet.

martinklepsch:
Some thoughts on populating the Lucene index: 

- We'll also want to add artifacts from Maven Central, this could happen via hardcoded list (https://github.com/cljdoc/cljdoc/issues/200) or through some other mechanism. I might do some work on making that extensible very soon.
- To abstract over the different repository sources it might make sense to move the data into a uniform location. The `releases` table might be a good candidate. It has previously been used to track releases and trigger new documentation builds but it seems within scope to contain a list of all relevant artifacts.
  - The `releases` table will need to be extended by a repository identifier, `clojars`, `maven-central`, etc.
  - There's a bit of overlap with the `versions` table but I think it's best to defer unifying this.
- The data from the `releases` could then be synced with the Lucene index periodically. Later on it could make sense to integrate the project's description but let's wait with that until  #208 is resolved.

holyjak:
FYI https://github.com/cljdoc/cljdoc/pull/306 has added a local search which could also be used to make the list of suggestions.

holyjak:
Replaced by #316

PR #225, opened 2018-11-20
show a warning when the build doesn't contain any namespace

### LINKED ISSUE #199: Warn when build succeeds but no namespaces/functions were found
In [build #5783](https://cljdoc.org/builds/5783) the analysis succeeded but didn't find any namespaces or functions. This can be a source of confusion since cljdoc claims everything went well and yet no namespaces will show up in the resulting page. To improve this we could do the following:

- [ ] add a boolean column to the builds table `has_namespaces` (or similar)
- [ ] set that column to `true` only if the [analysis result](https://6182-119377591-gh.circle-artifacts.com/0/cljdoc-edn/googlesheets-sql-sync/googlesheets-sql-sync/0.4.2/cljdoc.edn) contains any namespaces
- [ ] render a warning if the column is `false`

I will edit this issue after merging #197 and point to some places in the code that are relevant to implement this.

via #187

elarouss:
Fixes #199

### DIFF
--- a/src/cljdoc/render/build_log.clj
+++ b/src/cljdoc/render/build_log.clj
@@ -123,10 +123,12 @@
    (section
     (:api_imported_ts build-info)
     [:h3.mt0 "API Import"]
-    (if (or (nil? (:has_namespaces build-info))
-            (zero? (:has_namespaces build-info)))
+    (if (some-> build-info :namespaces_count zero?)
       [:p.bg-washed-red.pa3.br2.ma0 "No namespaces found"]
-      [:p.bg-washed-green.pa3.br2.ma0 "API imported successfully"])
+      [:p.bg-washed-green.pa3.br2.ma0
+       "Successfully imported "
+       (:namespaces_count build-info)
+       " namespaces"])
     [:p.mb0 (cljdoc-link build-info true)])])
 
 (defn build-page [build-info]
--- a/src/cljdoc/render/build_log.clj
+++ b/src/cljdoc/render/build_log.clj
@@ -123,7 +123,10 @@
    (section
     (:api_imported_ts build-info)
     [:h3.mt0 "API Import"]
-    [:p.bg-washed-green.pa3.br2.ma0 "API imported successfully"]
+    (if (or (nil? (:has_namespaces build-info))
+            (zero? (:has_namespaces build-info)))
+      [:p.bg-washed-red.pa3.br2.ma0 "No namespaces found"]
+      [:p.bg-washed-green.pa3.br2.ma0 "API imported successfully"])
     [:p.mb0 (cljdoc-link build-info true)])])
 
 (defn build-page [build-info]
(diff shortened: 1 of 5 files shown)


src/cljdoc/render/build_log.clj
```diff
@@ -123,7 +123,10 @@
    (section
     (:api_imported_ts build-info)
     [:h3.mt0 "API Import"]
-    [:p.bg-washed-green.pa3.br2.ma0 "API imported successfully"]
+    (if (or (nil? (:has_namespaces build-info))
```
martinklepsch: At this point `:has_namespaces` should always be set so let's maybe do something like
```
(some-> build-info :has_namespaces zero?)
```
the `some->` will achieve that `zero?` is only called if the `:has_namespaces` field is actually set.

martinklepsch: Or `(some-> build-info :imported_namespaces zero?)` if we decide to store the number of imported namespaces.

martinklepsch:
@elarouss looking good 👍  just  noticed the tests are failing, I think it might be this condition: https://github.com/cljdoc/cljdoc/blob/master/test/cljdoc/integration_test.clj#L60

You can run the tests locally with `clj -A:test`.

elarouss:
@martinklepsch i had that same issue the first time i ran tests (a week ago maybe), i thought maybe my internet connection is slow, so i changed : 
```clojure
(loop [i 20]
        (if (pos? i)
          (when-not (.contains (:body (pdt/response-for (service-fn @sys) :get build-uri))
                               "API imported successfully")
            (do (Thread/sleep 2000)
                (recur (dec i))))
          (throw (Exception. "Import took too long"))))
```
to this : 
```clojure
(loop [i 200] ;; yeah, a smart fix 
        (if (pos? i)
          (when-not (.contains (:body (pdt/response-for (service-fn @sys) :get build-uri))
                               "API imported successfully")
            (do (Thread/sleep 2000)
                (recur (dec i))))
          (throw (Exception. "Import took too long"))))
```
and then tests passed

martinklepsch:
Hey @elarouss, I think the issue is that the changes you made replace the `"API imported successfully"` with `"Successfully imported " (:namespaces_count build-info) " namespaces"`. The tests check for that "API imported successfully" string.

https://github.com/cljdoc/cljdoc/pull/225/files#diff-fc5c0e810687c2165acf3faae0ecfbd9L126

martinklepsch:
I pushed a fix to a branch: https://github.com/cljdoc/cljdoc/compare/elarouss-empty-ns-warning. 

Let me know if this looks reasonable to you and I'll merge it in 🙂

elarouss:
that's it :smile: tests are passing for me, you can merge it :v:

martinklepsch:
Merged, will deploy in a bit 👍 

I made another minor tweak adjusting the namespace count for when a namespace occurs in `clj` and `cljs` so it's not counted twice. Didn't think about this before: https://github.com/cljdoc/cljdoc/commit/93f5bf6ef888b0a6bc1316919f9190d0e7227964

martinklepsch:
It's live @elarouss 🎉

PR #227, opened 2018-11-21
Better separate Readme & Changelog from other articles #99

### LINKED ISSUE #99: Better separate Readme & Changelog from other articles
It'd be nice if there was some horizontal rule (or some other separator) to visually distinguish where the manual/articles for a project begin. Right now the README and the changelog are put in front (which makes a lot of sense), but they are not separated from the articles.

image

dawran6:
This addresses issue #99 

Fixes #99

### DIFF
--- a/src/cljdoc/render/articles.clj
+++ b/src/cljdoc/render/articles.clj
@@ -20,9 +20,10 @@
         " in your project."])])
 
 (defn main-list [doc-tree]
-  [:div.mb4.js--articles
-   (layout/sidebar-title "Main")
-   doc-tree])
+  (when doc-tree
+    [:div.mb4.js--articles
+     (layout/sidebar-title "Main")
+     doc-tree]))
 
 (defn doc-link [cache-id slugs]
   (assert (seq slugs) "Slug path missing")
--- a/src/cljdoc/render/articles.clj
+++ b/src/cljdoc/render/articles.clj
@@ -19,6 +19,11 @@
         [:a.blue.link {:href (util/github-url :userguide/scm-faq)} "properly set SCM info"]
         " in your project."])])
 
+(defn main-list [doc-tree]
+  [:div.mb4.js--articles
+   (layout/sidebar-title "Main")
+   doc-tree])
+
 (defn doc-link [cache-id slugs]
   (assert (seq slugs) "Slug path missing")
   (->> (string/join "/" slugs)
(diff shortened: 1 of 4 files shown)


dawran6:
Currently this PR increases the gap between Readme/Changelog and the rests.

image

dawran6:
It looks like this now: 
image

I'm not sure which is a better for this section, "MAIN" or "COMMON"? What do you think @martinklepsch ?

martinklepsch:
Nice that’s looking great 👍 I was thinking that we probably don’t even need the “Main” title. Maybe play around with dropping that and the indentation and see how it feels?

dawran6:
Sure! Let me work on that.

dawran6:
It looks a bit out of the place to me if I don't do anything with the font or padding.

image

It kind of breaks my expectation that "Readme" and "Changelog" should behave the same as the first indentation level (e.g. "ARTICLE" and "NAMESPACE") which isn't clickable nor blue.

Say if we change the font and padding to match the indentation level, should "ARTICLE" and "NAMESPACE" also be clickable?

image

dawran6:
Alt sidebar

martinklepsch:
I think I personally prefer the first variant with Readme and Changelog being styled as other articles. Perhaps with a bit more distance to the "ARTICLES" section heading.

Making the section headings ARTICLES and NAMESPACES clickable is an option but I consider them more visual spacers/section titles than something you'd interact with directly. Them not being blue (the color of things you can interact with on cljdoc) is also an issue. 

Unless you have strong feelings about this I'd suggest we go with the first iteration for now and reconsider this when there are more features coming up affecting this part of the UI.

dawran6:
Any code review would be appreciated. I feel the change I made in the article.clj can be dramatically improved tho I'm uncertain how.
image

src/cljdoc/render/articles.clj
```diff
@@ -19,6 +19,11 @@
         [:a.blue.link {:href (util/github-url :userguide/scm-faq)} "properly set SCM info"]
         " in your project."])])
 
+(defn main-list [doc-tree]
+  (when doc-tree
+    [:div.mb4.js--articles
```
martinklepsch: Please grep the codebase to ensure any JS code relying on this `.js--articles` class isn't negatively affected by this. I think there's only `doctree.js` which handles expanding of subtrees, which is not applicable here so it might be fine to just remove the class from this element.

dawran6:
I pushed a new commit based on your feedbacks @martinklepsch .
I ended up didn't change article-list but main-list because I need left padding (either `pl0` or `pl2`) on those `ul`s. The good part of that is we can apply `pl2` as-is in the doc-tree-view function. However, the drawback is that I replace the `pl2` with `pl0` in main-list function is not very future-proof. It might be broken if the style changes. Let me know what you think. Thanks!

bbatsov:
Nicely done!

PR #231, opened 2018-11-24
Track description and release date in versions table

### LINKED ISSUE #221: Integrate project description into versions table
Via #68 and also somewhat related to #208.

To expose better search functionality we will need not just group-id and artifact-id but also a project's description. In contrast to other contents of `pom.xml` this description needs to be stored in a searchable manner compared to the key-value style lookup described in #208.

#### Tasks

- [ ] Add a `description` column to the `versions` table that is set when a projects documentation is built
- [ ] Write a one-off script that retrieves descriptions and adds them to the `versions` table

Later on this data may be fed into a [FTS5](https://www.sqlite.org/fts5.html) SQLite table to enable full-text search based on description and other fields. [Customizing the weight](https://www.sqlite.org/fts5.html#the_bm25_function) of individual columns we should be able to provide reasonable search results. 

If you have thoughts on this, please chime in. 🙂

martinklepsch:
Fixes #221 

This adds `description` and `release_date` columns to the `versions` table. While this information is useful in multiple situations it's mostly intended for search related functionality. That is for situations where scanning the entirety of the data is important. 

In situations where we're only concerned with metadata on a single artifact I think the approach outlined in #208 is better suited.

This PR also includes some refactoring making `cljdoc.cli` use the same documentation build flow as the regular server.

### DIFF
--- a/modules/shared-utils/src/cljdoc/util.clj
+++ b/modules/shared-utils/src/cljdoc/util.clj
@@ -8,7 +8,9 @@
             [clojure.java.io :as io]
             [clojure.string :as string]
             [clojure.walk :as walk])
-  (:import (java.nio.file Files)))
+  (:import (java.nio.file Files)
+           (java.time Instant)
+           (java.time.format DateTimeFormatter)))
 
 (def hardcoded-config
   ;; NOTE `delay` is used here because the stripped-down analysis env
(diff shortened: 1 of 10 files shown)


martinklepsch:
Maybe adding this to the SQLite DB isn't right after all. SQLite's full text search capabilities seem nice but are fairly unergonomic compared to something more integrated like Lucene. And if our search isn't based on the SQLite tables then there is no point in putting this data there.

I'll pull out the refactoring made as part of this PR into a new PR.

PR #239, opened 2018-11-27
[WIP] Implement memoize and cache backed by SQL datastore

avichalp:
Provide an implementation of core.cache's `CacheProtocol`. This cache persists data on a SQL datastore. It is used with core.memoize's `PluggableMemoization` to provide memoizing functionality.


martinklepsch:
Hey! Sorry for not getting back to you about this earlier. The cache implementation looks good. I like that it’s using a separate databases that will allow us to exclude it from backups and similar.

I’m not so sure what I think of the Schema yet. Perhaps an extra prefix column would be fine?
As for migrations I think it would be cool if this was part of the library and not something the application would have to take care of.
I also meant to do something similar for the Clojars stats table (#233) but then decided against it. Might investigate that again.

avichalp:
Hey Martin,

Thanks for the comments.

You are right about migrations. I wanted to have the user of lib to pass the table name and column names rather than hard coding them. But JDBC doesn’t allow  dynamic table names in the query templates. So I had to do `format` on queries and I am not sure that if it is prone to sql injections. Although table name is not something that should come from client side but still I am not sure about this. What do you think?

Saw your ClojarStats PR do you think we can handle this with this caching system? Like we have a prefix something like `clojars-stats` and cache args and return values of a function that would fetch stats from Clojars?

Yes, I agree about having prefix in a separate column make sense I will change to that.

martinklepsch:
> Although table name is not something that should come from client side but still I am not sure about this. What do you think?

that was also what I bumped into, I think it should be fine as long as we expect the table name to be user supplied

also we could put some basic check in place, e.g. no spaces in the supplied value (edited)

> Saw your ClojarStats PR do you think we can handle this with this caching system? Like we have a prefix something like `clojars-stats` and cache args and return values of a function that would fetch stats from Clojars?

I was thinking about this but the stats system will have much more complex access patterns than just Key-Value that make it unsuitable for a similar implementation. For instance: 

- query 10 most downloaded libraries in the last 30 days
- query download count for specific library with custom interval / batching (i.e. per 7d/30d)

and so on. I'm not super happy that there is no reuse between these components but I think it's just too different 🙂

src/cljdoc/util/cache.clj
```diff
@@ -0,0 +1,122 @@
...
+  (has? [_ k]
+        (let [{:keys [spec key-prefix]} (:db state)
+              query "SELECT * FROM CACHE WHERE PREFIX = ? AND KEY = ?"]
+          (boolean
+           (value-> (sql/query spec [query key-prefix (pr-str k)])))))
+  (hit [_ k]
+       (SQLCache. state))
+  (miss [_ k v]
+        (let [{:keys [spec key-prefix]} (:db state)
+              value (nippy/freeze @v)
```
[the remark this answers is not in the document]
avichalp: Agreed. Good point. User should have the flexibility to pass in de/serialization functions.

avichalp:
Thanks for the review.

> It seems that all eviction related state is only kept in memory

Good point. Right now this cache implements only what we need for the memoize (`memo-sqlite`) function. For memoization purpose we don't need to evict anything. I can see that we need to refresh the cache when we add `TTL` functionality. For that we can do something similar to what `core.cache`'s TTL implementation does:
```clojure
  (has? [_ item]
    (and (let [[_ t] (get ttl item [0 (- ttl-ms)])]
           (< (- (System/currentTimeMillis)
                 t)
              ttl-ms))
         (contains? cache item)))
  (hit [this item] this)
  (miss [this item result]
    (let [now  (System/currentTimeMillis)
          [kill-old q'] (key-killer-q ttl q ttl-ms now)]
      (TTLCacheQ. (assoc (kill-old cache) item result)
                  (assoc (kill-old ttl) item [gen now])
                  (conj q' [item gen now])
                  (unchecked-inc gen)
                  ttl-ms)))
```
We can refresh the row without deleting it from table. If user still wants to evict something from the cache it is possible with [`core.memoize/memo-clear!`](https://github.com/clojure/core.memoize/blob/master/src/main/clojure/clojure/core/memoize.clj#L142-L158)  which calls `evict` on our cache that executes delete sql query.

To provide LRU-like functionality in this cache, I think we need to add another column for tracking last used timestamp of a row and also store a limit (size of LRU) somewhere. We also need to modify the implementation of `miss` to execute delete query on least recently used row when number of rows > size of cache. With this approach we won't have inconsistent state between memory and database. But I think this is beyond the scope this memoization and caching system we are trying to make. What do you think?

src/cljdoc/util/cache.clj
```diff
@@ -0,0 +1,123 @@
+(ns cljdoc.util.cache
```
martinklepsch: Just noticed: maybe `cljdoc.util.sqlite-cache` would be more explicit here?

src/cljdoc/util/cache.clj
```diff
@@ -0,0 +1,123 @@
...
+  "Memoizes the given function `f` using  `SQLCache` for caching.
+  `SQLCache` uses a SQL backend to store return values of `f`.
+
+  Example usage with SQLite:
+  ```
+  (def memo-f
+    (memo-sqlite (fn [arg] (Thread/sleep 5000) arg)
+                 {:key-prefix         \"artifact-repository\"
+                  :serialize-fn       identity
+                  :de-serialize-fn    read-string
```
martinklepsch: I think `:deserialize` might be the more common spelling here.

src/cljdoc/util/cache.clj
```diff
@@ -0,0 +1,123 @@
...
+
+  This namespace exposes `memo-sqlite` function which takes a
+  function to memoize and cache-spec, it retuns the memoized function.
+  "
+  (:require [clojure.core.cache :as cache]
+            [clojure.core.memoize :as memo]
+            [clojure.java.jdbc :as sql]
+            [taoensso.nippy :as nippy]))
+
+(cache/defcache SQLCache [state]
```
martinklepsch: You could make accessing of `cache-spec` fields easier by defining `SQLCache` like this:

```clojure
(cache/defcache SQLCache [db-spec key-prefix serialize-fn deserialize-fn])

(map->SQLCache {:db-spec ,,,
                :key-prefix "some-thing"
                :serialize-fn nippy/freeze
                :deserialize-fn nippy/thaw})
```

This way `serialize-fn` and so on could be accessed without having to retrieve it from the `state` map first.

> **Note:** I renamed `:db` to `:db-spec` in the code which matches more closely with the naming in cljdoc and `org.clojure/java.jdbc`.

martinklepsch:
Alright, let's roll with this for now then. I'm still a little sceptical as to whether `core.cache` really gets us anything with our usage but happy to go ahead with it for now and reconsider later if needed. 

I guess for initial integration a good step would be to incorporate this into the system defined in `cljdoc.server.system` and load the POM for a project into the request context using an interceptor. 

This could either be part of the `artifact-data-loader` below or a separate (but likely similar) interceptor.

https://github.com/cljdoc/cljdoc/blob/cf9f511b78b02667d097d84690a6c0960ac827d3/src/cljdoc/server/pedestal.clj#L95-L106

Eventually we'll want to integrate this into pages' description tag but that might be tricky at this point and I'd like to do some refactoring before I can knowingly send more contributors into those ruins 😄

avichalp:
As discussed earlier I have added  `TTL` functionality and migrations are now taken care of by the library. Also, user can pass the table name and column names that are used for storing cached values. Next I will be going forward with the integration part. I will start with caching contents of `pom` files as you suggested.

I think using `core.cache` helps to create this persistent memoization layer. With this solution I hope to tackle problems where we have to do multiple network calls to retrieve some information. With this in place the `memoized` function will be able to retrieve the same info by doing one sqlite query. Also since this data can be ad-hoc and not necessarily fit in our current schema. It might be good idea to put it as key, values in a secondary database. 

Really appreciate that you are willing to give it a chance. Lets see if this works out we expected. 🤞Otherwise can always it rip it out if it doesn't. 😅

src/cljdoc/util/cache.clj
avichalp: I changed the `:db` to `:db-spec`. After some implementing the `TTL` I had to do some refactoring now I need to pass the `cache-spec` to other functions so didn't change the field access part for now.

martinklepsch:
> As discussed earlier I have added TTL functionality and migrations are now taken care of by the library. 

I think we might have misunderstood each other here, I was only hypothetically hinting at the option of more elaborate caching schemes and didn't mean to suggest/request those to be implemented as part of this PR. Sorry for the confusion 🙊 

Feel free to keep it in for now but since there's no background job to evict old (beyond TTL) cache keys this will only cause occasional refreshing and won't limit the data actually stored.

In any case integrating this into actual responses and just getting it into production would be great and we can always reconsider the underlying mechanics later on. Integrating it via an interceptor would be a good start as mentioned earlier. Maybe integrating a projects description into the HTML `<meta description="...">` tag would be a good acceptance criteria for this PR.

If you fancy it you could also write up an ADR to preserve some of the context around the implementation for future contributors. Consider this optional though 🙂

src/cljdoc/util/sqlite_cache.clj
```diff
@@ -0,0 +1,188 @@
...
+  [{:keys [db-spec key-prefix table key-col value-col]}]
+  (let [column-specs [[:ttl "integer"]
+                      [:prefix "text"]
+                      [:cached_ts "text"]
+                      [(keyword key-col) "text"]
+                      [(keyword value-col) "text"]]
+        query (sql/create-table-ddl table
+                                    column-specs
+                                    {:conditional? true})]
+    (sql/execute! db-spec [query])))
```
martinklepsch: If the table is created like this there won't be any indexes except for row IDs — but our primary access is via the  `prefix` and `$key-col` columns. What I did in other tables and what is also appropriate here is to add a [`UNIQUE` constraint](https://www.techonthenet.com/sqlite/unique.php) to ensure there's no two rows with the same `prefix` and `$key-col`. This will implicitly add an index on the columns referenced as part of the `UNIQUE` constraint and therefore speed up queries significantly.

Also make sure that `prefix` and `$key-col` cannot be `null` since they are mandatory. Maybe even more columns should be enforced to have a value.

src/cljdoc/util/sqlite_cache.clj
```diff
@@ -0,0 +1,188 @@
...
+          (if (and (not (nil? item)) (stale? item))
+            (refresh! k v (:cache-spec state))
+            (cache! k v (:cache-spec state))))
+        (SQLCache. state))
+  (evict [_ k]
+         (evict! k (:cache-spec state))
+         (SQLCache. state))
+  (seed [_ base]
+        (seed! (:cache-spec base))
+        (SQLCache. base))
```
martinklepsch: Something is odd with indentation here, I think usually protocol method bodies are indented by two spaces:

```clj
  (seed [_ base]
    (seed! (:cache-spec base))
    (SQLCache. base))
```

I also think that `seed` maybe is not be the right place for the table creation, as I understand [the docstring](
https://github.com/clojure/core.cache/blob/8fd37130ccc574ce3c9e1de460808e7d628c1d81/src/main/clojure/clojure/core/cache.clj#L37-L40) it is more meant for putting some data into the cache which is not what's being done here. Not a big deal but felt like worth mentioning.

src/cljdoc/util/sqlite_cache.clj
avichalp: Good point about the indexes. I was about to ask you about that. As you suggested I have added a `Unique constraint` on (`prefix`, `$key-col`) also added `not nulls` on `prefix` , `$key-col` and `cached_ts`.

avichalp:
Thanks for the explanation. This makes a lot of sense. 🙂

In my new approach I have returned `cache` a map containing memoized functions from the `init` function.
Eg:
 ```clojure 
{:cljdoc.util.repositories/get-pom-xml (memo-sqlite cljdoc.util.repositories/get-pom-xml cache-spec)}
```
This is then passed to the interceptors. Now stateful components (memoized fns) are explicitly passed around. What do you think?

Also one question about description in `<meta>` tag.  Currently we a generic description like this:
https://github.com/cljdoc/cljdoc/blob/a8c7148e15592d2051bea2b99178a8218b6cc0b9/src/cljdoc/render/layout.clj#L25-L30

We can now add description of the artifact from pom file. So how do we change this message or should we add a new meta tag with project's description?

File: src/cljdoc/util/sqlite_cache.clj (hunk 6 above)
Comment (avichalp) [reply src/cljdoc/util/sqlite_cache.clj]: Fixed indent.
Yes, fair point about the `seed` fn. To me it looks like this should be used to put initial data. But our implementation requires table creation before we seed cache with any data. Thats why I added table creation in `seed`. I not sure of any better place to put table creation logic at the moment. May be we can leave it for now and evaluate later or if you have any suggestion I can implement that?

**Comment by martinklepsch** [comment]
> In my new approach I have returned cache a map containing memoized functions from the init function.

Looks good 👍 Integrant init methods can also return functions but the map approach is equally fine of course. 

> So how do we change this message or should we add a new meta tag with project's description?

I'm not sure. Perhaps something like the following could work?

```clj
(str artifact-id ": " description ". Documentation for %s v%s on cljdoc.")
```

src/cljdoc/renderers/html.clj
```diff
@@ -101,7 +111,11 @@
                        :canonical-url (some->> (bundle/more-recent-version cache-bundle)
                                                (merge route-params)
                                                (routes/url-for :artifact/namespace :path-params))
-                       :description (layout/description cache-id)}))))
+                       :description (layout/artifact-description
+                                     cache-id
+                                     (-> (Jsoup/parse (-> cache-contents :version :pom))
+                                         (.select "description")
+                                         (.text)))}))))
```
martinklepsch: Using `cljdoc.util.pom` has some functions for this that could be used here.

src/cljdoc/server/system.clj
```diff
@@ -66,6 +71,15 @@
 (defmethod ig/init-key :cljdoc/dogstats [_ {:keys [uri tags]}]
   (dogstatsd/configure! uri {:tags tags}))
 
+(defmethod ig/init-key :cljdoc/cache [_ {:keys [cache-dir]}]
+  (.mkdirs (io/file cache-dir))
+  {:cljdoc.util.repositories/get-pom-xml (sqlite-cache/memo-sqlite
```
martinklepsch: `init-key` can also return plain functions so you could just return the memoized function making usage a little more concise in other places.

src/cljdoc/server/system.clj
```diff
@@ -6,13 +6,16 @@
...
             [clojure.tools.logging :as log]
             [clojure.java.io :as io]
             [cognician.dogstatsd :as dogstatsd]
             [integrant.core :as ig]
             [unilog.config :as unilog]
             [ragtime.jdbc :as jdbc]
-            [ragtime.core :as ragtime]))
+            [ragtime.core :as ragtime]
+            [taoensso.nippy :as nippy]
+            [cljdoc.util.sqlite-cache :as sqlite-cache]))
```
martinklepsch: No big deal but this require could be moved closer to the other `cljdoc.*` requires :)

martinklepsch:
@avichalp I just added a few more minor comments but otherwise would be happy to get this merged. Are you happy with the state of the PR? 

One thing that might be still nice is a small test that verifies that the project's description is part of a projects page (integration test).

src/cljdoc/server/system.clj
avichalp: Thats a good idea but I wanted to return a `map` because we can use this `map` to add more memoized functions in future. For ex: may be we can have something like `{... :cljdoc.util.repositories/get-github-stars (sqlite-cache/memo-sqlite ...)}` etc. 

Is this the right way to think? or am I missing something?

avichalp:
> One thing that might be still nice is a small test that verifies that the project's description is part of a projects page (integration test)

Good idea. I will be adding an integration test in next commits. I think it would be to merge after that.

src/cljdoc/server/system.clj
martinklepsch: I think in this case it would be more idiomatic to just add the additional caches as other top level items in the integrant system map?

Issue #184, opened 2018-10-24
Automatically deploy on commits to master?


danielcompton:
It would be nice if commits to master automatically triggered deploys. As a contributor, it's always motivating to see your work immediately in production.

One of the issues with this will be the downtime when doing a deploy. I'm not sure how much of this can be reduced by doing things like building an Uberjar. If you had a CDN/Nginx cache sitting in front of the application then you could probably reduce the downtime further still.

You can also do some [cool stuff](https://developer.github.com/v3/guides/delivering-deployments/) with the GitHub API around handling deployments. See https://github.com/rubytogether/rubytogether.org/pull/213#event-1722187996 for what this can look like.

martinklepsch:
> If you had a CDN/Nginx cache sitting in front of the application then you could probably reduce the downtime further still.

We do have Nginx in front of the app. Do you envision a setup where a second service is started and switched to once available? Or more like Nginx serving cached responses while the actual service is down? 

Either of these would be great improvements... 😃

martinklepsch:
I looked a bit into how this could be achieved today. Nginx can handle this by running old and new config side by side for a bit:

> NGINX configuration is captured in configuration files. When you change a configuration file and restart NGINX to pick up the new configuration, it implements a “graceful restart”. Both the old and new copies of NGINX run side by side for a short period of time. The old processes don’t accept any new connections and terminate once all their existing connections terminate. 

I also stumbled upon [Caddy](https://caddyserver.com) which supports a [upstream directive](https://caddyserver.com/docs/proxy) that can take a port range:

> **try_duration** is how long to try selecting available upstream hosts for each request. By default, this retry is disabled ("0s"). Clients may hang for this long while the proxy tries to find an available upstream host. This value is only used if a request to the initially-selected upstream host fails.

 Another nice thing about Caddy is that it takes care of SSL certificates automatically (and perhaps is easier to test due to it's single config file approach). [Traefik](https://traefik.io) also has built-in Lets Encrypt support but seems a little more involved/complex for our use case.

--- 

I'm still not sure what the best way to cycle cljdoc processes would be.

Presumably upon deploy we would start another cljdoc process on the first free port in a given range, wait for it to become available and then terminate the existing (old) process. 

How to achieve that with systemd isn't quite apparent to me yet.

martinklepsch:
Closed via https://github.com/cljdoc/cljdoc/commit/970c543e13bea78cd9b2c75d7563e36b85b54610 🎉

PR #248, opened 2018-12-03
cljdoc ops 2.0

martinklepsch:
Initial groundwork for #184 and general revisit of cljdoc's devops tooling.

In short this PR:

- Defines a new CentOS based image with [Nomad](https://www.nomadproject.io/), [Consul](https://www.consul.io) and Docker.
- Adds Clojure code to deploy cljdoc to Nomad via it's [HTTP API](https://www.nomadproject.io/api)
- Uses [Traefik](https://traefik) as reverse proxy and for service discovery via Consul
- Implements a zero-downtime swap when deploying new versions of cljdoc

Take a look at the new readme [`ops/README.adoc`](https://github.com/cljdoc/cljdoc/blob/blue-green/ops/README.adoc), the [code responsible for deploying to Nomad](https://github.com/cljdoc/cljdoc/blob/blue-green/modules/deploy/src/cljdoc/deploy.clj) and the [decision record](https://github.com/cljdoc/cljdoc/blob/blue-green/doc/adr/0017-use-nomad-for-deployment.md) for this change.


danielcompton:
The ADR explains the motivations well. I do worry a little bit about how much extra complexity this will add. The nginx + bash scripts route has the advantage of being easier up front as it is made up of things people understand well, though you make a good point about bit-rot.

Most of cljdoc is read-only, if there was a caching CDN (or even just nginx) in front that could serve URLS out of the cache, that would seem to take away most of the downsides of more frequent deployments, though it wouldn't be as complete of a solution as what you're describing here.

.circleci/config.yml
```diff
@@ -85,12 +81,27 @@ jobs:
...
+
 workflows:
   version: 2
   build-and-deploy:
     jobs:
       - build
       - prettier
+      - docker-deploy:
+          requires:
+            - build
```
danielcompton: This needs a branch filter like deploy I think?

ops/infrastructure/cljdoc.tf
```diff
@@ -65,6 +78,15 @@ resource "aws_route53_record" "dokku" {
   records  = ["167.99.133.5"]
 }
 
+resource "aws_route53_record" "test_xyz" {
```
danielcompton: We probably don't need to add a record for test.cljdoc.xyz, if it's never been used before?

ops/nomad.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
```
danielcompton: If this was sourced from somewhere else it could be good to reference it here so people understand any extra context.

ops/nomad.sh
```diff
@@ -0,0 +1,40 @@
...
+
+trap cleanup EXIT ERR INT TERM
+
+# Start SSH port forwarding process for consul (8500) and nomad (4646)
+ssh -M -S ${socket} -fNT -L 8500:localhost:8500 -L 4646:localhost:4646 root@${ip}
+
+ssh -S ${socket} -O check root@${ip}
+
+echo "You can now open Nomad or Consul:"
+echo "Nomad: http://localhost:4646/"
```
danielcompton: Are these ports exposed publicly? Doesn't look like it, but thought I'd check.

danielcompton:
I mostly just clicked the request changes button to see if they'd changed the icon, as the UI looked a little different :). This seems pretty reasonable after looking over the code and seeing the explanation. It's a little bit terrifying seeing all of the extra things coming along for the ride (Nomad, Consul, Traefik), but I think the end goal of automatically deploying to production is a worthy one.

One other thought I had is about the database. It has been a bit awkward to do this kind of work because we have a stateful sqlite db on the application server. If we had Postgres on a separate server then we could easily recycle app servers without losing any state. Though that adds it's own set of complexity... This seems pretty good to me.

.circleci/config.yml
martinklepsch: The used Docker tag contains the branch name if it isn't `master` but yeah, we probably just shouldn't push to Docker unless we're on `master`.

ops/infrastructure/cljdoc.tf
martinklepsch: That's right, I used this for testing mostly, will removed again eventually.

ops/nomad.sh
martinklepsch: They are forwarded from the production host as part of the script.

ops/nomad.sh
danielcompton: Sorry, "sourced" was a bad term here, I meant if you had copied the script from elsewhere then it would be good to add the URL here. But also just a docstring on the purpose would be good too.

martinklepsch:
> It's a little bit terrifying seeing all of the extra things coming along for the ride (Nomad, Consul, Traefik)

I was equally terrified when I started going down this road but especially Nomad and Consul have proven very predictable and are exceptionally well documented. I'm still a little bit worried about Traefik as it seems to have various [minor](https://github.com/containous/traefik/issues/4247) bugs but it's working for now. Various people mentioned [fabio](https://fabiolb.net) but it doesn't handle SSL certs automatically and so we'd loose some of the niceties of Traefik in that regard.

> One other thought I had is about the database. 

This came into my mind a few times while working on this too. I still very much like the single instance setup that we have now (i.e. avoiding distributed systems problems) but if we later decide we want a separate instance with a database the Nomad/Consul setup will only make this easier.

 > The nginx + bash scripts route has the advantage of being easier up front as it is made up of things people understand well, though you make a good point about bit-rot.

Besides bit-rot I also found this setup much easier to test. Basically you just add another `server_instance`:

```hcl
module "test_server" {
  source     = "./server_instance"
  # ...
}
```

And then run `cljdoc.deploy` against the IP of this instance. This is what I did with the `test.` instance and it allows devs to spin up a new environment in a production setting (SSL and all) in a matter of minutes.

---
**PS.** I realize that a lot of what I'm writing is fairly argumentative despite not having to convince anybody. As you might be guess I'm fairly invested in this particular approach just by the sheer amount of time I spent on it. If you (not just Daniel) think this is going in the wrong direction, make yourself heard please. I'm also available on Slack as @martinklepsch.

ops/nomad.sh
martinklepsch: That `ssh -S /tmp/some-socket` magic is very nice if you ever want to forward some ports in the background. There's more details on it here: http://mpharrigan.com/2016/05/17/background-ssh.html

ops/nomad.sh
martinklepsch: Done 👍

martinklepsch: And some more on the what the thing is actually for: https://github.com/cljdoc/cljdoc/pull/248/commits/01e6cbe9e8198723bca5e6446cc6febe2c0121eb

danielcompton:
> PS. I realize that a lot of what I'm writing is fairly argumentative despite not having to convince anybody. As you might be guess I'm fairly invested in this particular approach just by the sheer amount of time I spent on it.

Just saw this now, I missed your earlier comment. I didn't take your comments to be argumentative, just explaining clearly why you went this route. I think it's a good choice, I'm excited to see it get into prod!

Issue #257, opened 2019-01-15
Render Cucumber "features" as articles


plexus:
Cucumber is a multi-platform testing tool, when using Cucumber you write your tests in a language called "Gherkin", which looks like lightly structured plain text. Each test file is known as a "feature", and uses the `.feature` extension (these are meant to be "integration" tests, so supposedly each file describes a single feature of your library/app).

Here's an [example of a feature file](https://github.com/lambdaisland/kaocha/blob/master/test/features/pending.feature). This can be executed as a test (provided you provide definitions of the "steps" used, [here's an example of that](https://github.com/lambdaisland/kaocha/blob/master/test/step_definitions/kaocha_integration.clj#L172-L174)). 

The benefit of writing your tests in this roundabout way is that your tests are now optimized for being read by humans, and so can straightforwardly be rendered as HTML or Markdown. Here's the same "test" [rendered as a cljdoc article](https://cljdoc.org/d/lambdaisland/kaocha/0.0-359/doc/marking-tests-as-pending).

To try this out I've generated the markdown and checked it into the repo, but now I have generated files checked in that I need to manually keep up to date with the source files. If cljdoc could understand `.feature` files the way it understands `.md` or `.adoc` files this would be much more straightforward, and allow for nicer rendering of the output, generating a table of contents, etc.

If this is considered useful I'd be happy to contribute support for it. I already did the work of wrapping `cucumber-jvm`, so it's easy to parse a `.feature` file to a plain EDN data structure.

I'm not 100% sure yet what the behavior should be regarding detecting `.feature` files, and combining them with other articles, but I can follow up with a more concrete proposal to discuss.

Note that this could be considered related to #70, but it's not really the same thing, so I'd rather not pull that whole discussion into this one. It could be a good starting point though, e.g. with this in place we could think about ways to make examples in Cucumber tests interactive.

plexus:
That's right, but it's no different from how articles are currently rendered.

plexus:
The easiest first step would be to make it opt-in, by referencing each feature file explicitly from `cljdoc.edn`

``` edn
{:cljdoc.doc/tree
 [["Do this thing" {:file "test/features/do_this_thing.feature"}]]}
```

That seems like a fairly uncontroversial starting point that should be easy to implement.

Projects that really embrace Cucumber as documentation on the other hand will have a structure like this

```
features
├── README.md
├── command_line_args
│   ├── README.md
│   ├── no_color.feature
│   └── fail_fast.feature
[...]

PR #262, opened 2019-01-28
Extensible article types

martinklepsch:
This PR opens up article types to be extended beyond Markdown and Asciidoc. This will be useful for experimental extensions like Cucumber feature files as described in #257 as well as proprietary formats used inside organisations with an on-premise cljdoc deployment. 

One aspect I'd particularly appreciate feedback about is the whole `filepath->type` stuff. Basing this purely on the extension seems a little too constrained and the most appropriate approach would probably be a sequence of predicates. That's not as nice as multi methods however and would require some manual state-wrangling in the form of an atom or similar. Wouldn't be the end of the world, multimethods are just state anyways but still, multimethods are more familiar and very well integrated into the language.

### DIFF
--- a/doc/custom-articles.adoc
+++ b/doc/custom-articles.adoc
@@ -26,6 +26,6 @@ customize the configuration of cljdoc's `config.edn`:
 [source,clojure]
 ----
 {,,,
- :extension-namnespaces [com.yourproject.cljdoc-articles]}
+ :extension-namespaces [com.yourproject.cljdoc-articles]}
 ----
 
--- a/doc/custom-articles.adoc
+++ b/doc/custom-articles.adoc
@@ -1,0 +1,31 @@
+= Custom Articles
+:toc:
+
+cljdoc supports Markdown and Asciidoc out of the box but support for additional content types can be added in one of the following ways:
+
+1. Extending the cljdoc codebase itself to cover a new content type
+1. Packaging a separate jar that contains the necessary code to support additional content types
+
+The actual code will mostly be the same for both approaches, packaging and integration however will be different.
+
+*What approach to take?* If you think the content type could be beneficial to the wider cljdoc userbase it can be a part of cljdoc itself. If you want to use cljdoc on-premise in your company and have custom content types not used outside your company packaging them separately is probably the best way.
+
+== Extending cljdoc
+
+At the core adding support for new filetypes happens by extending two multimethods:
+
+- `cljdoc.doc-tree/filepath->type` is used to infer what content type a file contains based on it's path in a Git repository. For `.markdown` this function returns `:cljdoc/markdown`.
+- `cljdoc.rich-text/render-text` is used to turn the actual content into HTML.
+
+== Packaging Externally
+
+The above multimethods can also be extended from third party jars. To ensure that namespaces
+from additional jars are loaded (and thus those multimethods are extended as expected) you can
+customize the configuration of cljdoc's `config.edn`:
+
+[source,clojure]
+----
+{,,,
+ :extension-namnespaces [com.yourproject.cljdoc-articles]}
+----
+
--- a/src/cljdoc/config.clj
+++ b/src/cljdoc/config.clj
@@ -104,6 +104,9 @@
 (defn maven-repositories []
   (get-in (config) [:maven-repositories]))
 
+(defn extension-namespaces [config]
+  (clojure.core/get-in config [:extension-namespaces]))
+
 (comment
   (:cljdoc/server (config))
 
--- a/src/cljdoc/doc_tree.clj
+++ b/src/cljdoc/doc_tree.clj
@@ -58,26 +58,43 @@
              :attrs (spec/? ::hiccup-attrs)
              :children (spec/* ::hiccup-entry))))
 
+(defmulti filepath->type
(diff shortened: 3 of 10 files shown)


doc/custom-articles.adoc
```diff
@@ -0,0 +1,31 @@
+= Custom Articles
+:toc:
+
+cljdoc supports Markdown and Asciidoc out of the box but support for additional content types can be added in one of the following ways:
+
+1. Extending the cljdoc codebase itself to cover a new content type
+1. Packaging a separate jar that contains the necessary code to support additional content types
```
avichalp: s/1./2./

martinklepsch: This kind of numbering is fortunately handled by Markdown and Asciidoc automatically :)

src/cljdoc/config.clj
```diff
@@ -104,6 +104,9 @@
 (defn maven-repositories []
   (get-in (config) [:maven-repositories]))
 
+(defn extension-namespaces [config]
+  (clojure.core/get-in config [:extension-namespaces]))
```
avichalp: May be just `get-in`? Why do we use a fully qualified name here?

martinklepsch: The namespace defines it's own `get-in` which throws if a value is missing. It also excludes `clojure.core/get-in` from being referred alongside the rest of `clojure.core`:

```clj
(ns cljdoc.config
  (:refer-clojure :exclude [get-in])
```

avichalp: ah! understood. I should have read the code above. Thanks for clarification 🙂

doc/custom-articles.adoc
```diff
@@ -0,0 +1,31 @@
...
+== Packaging Externally
+
+The above multimethods can also be extended from third party jars. To ensure that namespaces
+from additional jars are loaded (and thus those multimethods are extended as expected) you can
+customize the configuration of cljdoc's `config.edn`:
+
+[source,clojure]
+----
+{,,,
+ :extension-namnespaces [com.yourproject.cljdoc-articles]}
```
plexus: Typo: `s/namne/name/`

src/cljdoc/render/rich_text.clj
```diff
@@ -70,6 +70,19 @@
...
+
+  `type` is determined by [[cljdoc.doc-tree/filepath->type]]."
+  (fn [[type contents]]
+    type))
+
+(defmethod render-text :cljdoc/markdown [[_ content]]
+  (markdown-to-html content))
+
+(defmethod render-text :cljdoc/asciidoc [[_ content]]
+  (markdown-to-html content))
```
plexus: `s/markdown/asciidoc/` ?

src/cljdoc/doc_tree.clj
```diff
@@ -139,29 +169,24 @@
...
+       (try (filepath->type path) (catch Exception _ false))))
 
 (defn- infer-title [path file-contents]
-  (or (case (supported-file-type path)
-        :markdown (second (re-find #"(?m)^\s*#+\s*(.*)\s*$" file-contents))
-        :asciidoc (second (re-find #"(?m)^\s*=+\s*(.*)\s*$" file-contents)))
+  (or (case (filepath->type path)
+        ;; NOTE infer-title will fail with non adoc/md files
+        :cljdoc/markdown (second (re-find #"(?m)^\s*#+\s*(.*)\s*$" file-contents))
+        :cljdoc/asciidoc (second (re-find #"(?m)^\s*=+\s*(.*)\s*$" file-contents)))
```
plexus: Could this also benefit from a multimethod?

plexus:
I think this looks pretty cool, and would certainly make it easier for me to do a first iteration fo Cucumber feature rendering. I left a few remarks inline.

One things that's not immediately clear from reading the code: can I override the file type from `cljdoc.edn`. Say I have some Markdown files that I actually want to see rendered in a special way, say add some pre-processing, can I do something like this?

``` clojure
{:file "doc/README.md"
 :cljdoc.doc/type :my.special/markdown}
```

Thanks!

src/cljdoc/doc_tree.clj
martinklepsch: It could but I don't want to put a lot more stuff into the infer-logic, people should use `doc/cljdoc.edn` and cljdoc should provide appropriate tooling to generate those files, verify them, etc.

martinklepsch:
``` clojure
{:file "doc/README.md"
 :cljdoc.doc/type :my.special/markdown}
```
You can't but that might actually be a good idea. It's trivial to implement but I'll intentionally skip it for v1 of this. Every feature has a cost and so on :) If you ever actually need it, just let me know and we'll make it happen 😄 

Thanks for the other comments, greatly appreciated and you caught some bugs 🙌

plexus:
> If you ever actually need it, just let me know and we'll make it happen smile

I was thinking of perhaps using it for the table of contents I talked about in #257. Then again, I might not. Still seems like a good idea if you're guessing/deriving values to allow people to bypass the guessing if they know what they want.

Another question, from the docs it sounds like for these custom types to work on cljdoc.org they would have to be contributed to cljdoc itself, but would it not also work if they're implemented in the codebase whose docs are being generated? That would mean a lot more flexibility.

PR #264, opened 2019-02-01
Improved support for modular libraries

### LINKED ISSUE #24: Allow inclusion of namespaces provided by dependencies in API docs
[Reitit](https://github.com/metosin/reitit), [Yada](https://github.com/juxt/yada/) and others are composed of multiple smaller libraries. Documentation often refers to namespaces within those libraries, which are not part of the actual "main artifact".

It would be nice to better support these projects by allowing them to include API docs for namespaces in their dependencies in some way.

A potential approach could be introducing a configuration option:

```clj
;; doc/cljdoc.edn
{:cljdoc/include-namespaces-from-dependencies [reitit/core reitit/ring ,,,]
```

This information could be used to retrieve namespaces from those dependencies and add them to the list of namespaces in the sidebar. 

Things to watch out for:
1. We'll need to make sure we include the right version of this namespace (i.e. look in the main artifacts `pom.xml`)
2. The UI should indicate that these namespaces come from a different artifact
3. yada provides different bundles, we should only include namespaces that the individual bundle depends upon

**EDIT:** Ring also is this kind of library.

### LINKED ISSUE #198: Allow doc/cljdoc.edn to address multiple artifacts
Since having multiple modules in single Git repositories is becoming a thing (see [reitit](https://github.com/metosin/reitit/tree/master/modules) & [muuntaja](https://github.com/metosin/muuntaja/tree/master/modules)) the `cljdoc.edn` configuration file should allow users to supply different configuration<sup>1</sup> for each module/artifact.

<sup>1</sup> Currently the only configuration is doc trees but in the future this may be expanded to analysis configuration like excluding side-effecting files from analysis or inlined dependencies as in `nrepl.inlined-deps.*`

#### Proposal

Currently the map in `doc/cljdoc.edn` is expected to have various keywords at the top level, most notably `:cljdoc.doc/tree`. In order to maintain this we can allow additional symbol keys in the EDN map whose respective values will take precedence if the symbol matches the project currently analysed. An example:

```clj
{:cljdoc.doc/tree [["Overview" {:file "doc/overview.md"}]]
 metosin/reitit {:cljdoc.doc/tree [["Introduction" {:file "intro.md}]]}}
```
With a config like this `Introduction` will only be rendered for `metosin/reitit` while `Overview` will be rendered for every other project that points to the same Git repository. 

This is backwards compatible and keeps the configuration format straightforward for those who don't maintain multiple projects in a single Git repository.

#### Feedback

I would appreciate feedback on this proposal especially from those who maintain multi-module repositories. (:wave: @ikitommi)

Related to #24

martinklepsch:
Fixes #24 
Fixes #198 

This PR significantly improves cljdoc's usability with regards to libraries consisting of multiple modules/separate jars. Notably the following features are added:

#### `doc/cljdoc.edn` can target multiple artifacts

If a repository contains the code for multiple artifacts it might be desirable to provide different doc trees for those artifacts. This can be used to point users to the "ubermodule" documentation or to provide more detailed documentation on a module.

#### Namespaces from dependencies can be included

Ubermodule libraries often don't contain any actual code and rely on dependencies to import functionality. Users can now provide a `:cljdoc/include-namespaces-from-dependencies` key to include APIs from dependencies in the documentation of an artifact.

💡  For details on how to use those features, check the [changes to the user guide](https://github.com/cljdoc/cljdoc/blob/multi-module-docs/doc/userguide/for-library-authors.adoc#module-support) part of this PR.

🗣 As usual, feedback is very welcome :)

screenshot 2019-02-01 at 20 25 49

### DIFF
--- a/modules/shared-utils/src/cljdoc/util.clj
+++ b/modules/shared-utils/src/cljdoc/util.clj
@@ -69,7 +69,7 @@
 (defn git-dir [project version]
   (str "git-repos/" (group-id project) "/" (artifact-id project) "/" version "/"))
 
-(defn clojars-id [{:keys [group-id artifact-id] :as cache-id}]
+(defn clojars-id [{:keys [group-id artifact-id] :as artifact-entity}]
   (if (= group-id artifact-id)
     artifact-id
     (str group-id "/" artifact-id)))
--- a/modules/shared-utils/src/cljdoc/util.clj
+++ b/modules/shared-utils/src/cljdoc/util.clj
@@ -238,3 +238,6 @@
     (let [sqr  (fn sqr [x] (* x x))
           avg  (mean coll)]
       (mean (map #(sqr (- % avg)) coll)))))
+
+(defn index-by [f m]
+  (into {} (map (juxt f identity) m)))
--- a/src/cljdoc/storage/api.clj
+++ b/src/cljdoc/storage/api.clj
@@ -27,7 +27,7 @@
                             (:artifact-id version-entity)
                             [(:version version-entity)])
     (sqlite/import-api db-spec version-entity codox))
-  (import-doc [_ version-entity {:keys [doc-tree scm jar] :as version-data}]
+  (import-doc [_ version-entity {:keys [doc-tree scm jar config] :as version-data}]
     (sqlite/import-doc db-spec version-entity version-data))
   (exists? [_ version-entity]
     (sqlite/docs-available? db-spec
--- a/deps.edn
+++ b/deps.edn
@@ -56,9 +56,7 @@
          "modules/shared-utils/resources"]
  :aliases {:test
            {:extra-paths ["test"]
-            :extra-deps {lambdaisland/kaocha
-                         {:git/url "https://github.com/lambdaisland/kaocha.git"
-                          :sha "e7af8b48aadb32da9dcd7a9e5eb50fe64eae8d62"}}
+            :extra-deps {lambdaisland/kaocha {:mvn/version "0.0-389"}}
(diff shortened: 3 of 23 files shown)


src/cljdoc/storage/api.clj
```diff
@@ -27,7 +27,7 @@
                             (:artifact-id version-entity)
                             [(:version version-entity)])
     (sqlite/import-api db-spec version-entity codox))
-  (import-doc [_ version-entity {:keys [doc-tree scm jar] :as version-data}]
+  (import-doc [_ version-entity {:keys [doc-tree scm jar config] :as version-data}]
```
avichalp: Looks like `config` and other keys are not being used. Why not skip destructuring?

modules/shared-utils/src/cljdoc/util.clj
```diff
@@ -238,3 +238,6 @@
     (let [sqr  (fn sqr [x] (* x x))
           avg  (mean coll)]
       (mean (map #(sqr (- % avg)) coll)))))
+
+(defn index-by [f m]
+  (into {} (map (juxt f identity) m)))
```
avichalp: So cool! 😎

src/cljdoc/storage/api.clj
martinklepsch: To remind myself what's in that map. Honestly it's probably a stupid thing to do as it will inevitably get out of date.

modules/shared-utils/src/cljdoc/util.clj
martinklepsch: Stolen from other smart people 😎

Issue #206, opened 2018-11-13
Get cljdoc on cljdoc


martinklepsch:
cljdoc isn't a library but it might still be nice to have it's API documented on cljdoc:

- we could set a standard for good API documentation as well as usage of more advanced features like articles, wikilinks and more
- it could be useful for contributors to familiarise themselves with the various components of cljdoc
- like self-hosting for ClojureScript, it's simply the right thing to do 👌 

Things needed to make this happen:

- [x] package cljdoc as a jar
- [ ] publish that jar to Clojars

Packaging a jar is mostly solved with pack.alpha, depstar or similar but publishing things to Clojars and generating the required `pom.xml` is still somewhat hacky (you have to generate it using `clj -Spom` and then make various edits). 

I'd like the end result to be as simple as this:

```
clj -m something.publish --project cljdoc --version 0.1.0
```

timgilbert:
Hey, so I just published https://github.com/workframers/garamond which does some pom.xml post-processing, including creating or updating the `<scm>` tag. It's still brand-new and I hope to write a little blog summary, but the README is decent and it is able to deploy itself to clojars from circleci without needing to check a pom.xml into source control. Check it out!

martinklepsch:
@elarouss Hey there! :) I remember you worked on this and you seemed to have made some progress, would be awesome if you could open a PR 👍

martinklepsch:
BAM! http://cljdoc.org/d/cljdoc — Thanks @elarouss for his work on this.

PR #265, opened 2019-02-02
deploy cljdoc using garamond and deps-deploy

elarouss:
To deploy, first generate the pom file using garamond: 
`$ clojure -A:garamond -p` 
and then deploy using deps-deploy: 
`$ clojure -A:deploy`

Part of #206


martinklepsch:
Made some tweaks and merged this manually: https://github.com/cljdoc/cljdoc/commit/43dbc8dee605dfa2a135092608b992912370742d

Thanks for looking into this, will be exciting to have cljdoc on cljdoc. :)

Issue #273, opened 2019-02-09
Local JARs and classpath


residentsummer:
I was trying to ingest a locally-built jar with a custom version string into cljdoc storage and encountered a bug (as I think).

I've found out that local jar isn't used for building the classpath. Instead, a maven dep is added to the deps map (https://github.com/cljdoc/cljdoc/blob/master/modules/analysis-runner/src/cljdoc/analysis/deps.clj#L92), pointing to project@version. So, after all manipulations the analysis impl got project classes on CP and an unpacked jar via command-line args. Both groups of files _may_ be of different versions.

I've tried a few approaches to overcome this issue, with mixed results:

1. First, I thought that jar path gets to runner_ng only if specified via CLI, so I've added a conditional expr to replace :mvn/version with :local/root if jar url is provided. This worked, but required changes seemed too intrusive for me (passing jar-uri through many funcs down to `deps/deps`. Besides, I wasn't sure, that including a project itself in the output of `(deps pom)` is conceptually right. And, as I known later, jar path is always there...
2. Second, I've removed `(assoc ... project {:mvn/version version})` from `deps/deps` and added a path to unpacked jar to the CP in `runner/analyze-impl`. Surprisingly, this didn't worked. I've found the root cause to be the way how `deps/deps` and `deps/extra-deps` work - only provided, test and system dependencies are included without explicit project dependency. I was curious about it and found this - https://github.com/cljdoc/cljdoc/issues/147. Since there us no equivalent of `<type>pom</type>` in tools.deps, this is a dead-end.
3. With all this knowledge, I've merged (1) and (2) successfully. The only issue that remained is whether to 1) leave project jar on CP and analyze unpacked files from different location 2) source both from one place. I went with the second option.

By the way, why do we need project files on CP? Is this a requirement for codox?

I've made a preliminary patch for this - could you please take a look? Or, maybe, I've got something terribly wrong with the whole situation :)

residentsummer:
```
diff --git a/modules/analysis-runner/src/cljdoc/analysis/deps.clj b/modules/analysis-runner/src/cljdoc/analysis/deps.clj
index 56c9451..d88cabb 100644
--- a/modules/analysis-runner/src/cljdoc/analysis/deps.clj
+++ b/modules/analysis-runner/src/cljdoc/analysis/deps.clj
@@ -88,8 +88,7 @@
         (merge (get hardcoded-deps project))
         (ensure-required-deps)
         (ensure-recent-ish)
[...]

PR #276, opened 2019-02-12
Improvements for ingesting local JARs

residentsummer:
* use it for building CP (#273)
* use artifact version from analysis result instead of pom's one

### DIFF
--- a/modules/analysis-runner/src/cljdoc/analysis/deps.clj
+++ b/modules/analysis-runner/src/cljdoc/analysis/deps.clj
@@ -106,8 +106,7 @@
   {:pre [(string? pom-url)]}
   (println pom-url repos)
   (let [pom (pom/parse (slurp pom-url))
-        {:keys [group-id artifact-id version]} (pom/artifact-info pom)
-        project (symbol group-id artifact-id)
+        project (util/clojars-id (pom/artifact-info pom))
         resolved (tdeps/resolve-deps {:deps (deps pom),
                                       :mvn/repos (merge default-repos
                                                         (extra-repos pom)
--- a/modules/analysis-runner/src/cljdoc/analysis/deps.clj
+++ b/modules/analysis-runner/src/cljdoc/analysis/deps.clj
@@ -103,7 +103,7 @@
 (defn resolved-and-cp
   "Build a classpath for the project specified by `pom-url`."
   [jar-url pom-url repos]
-  {:pre [(string? pom-url)]}
+  {:pre [(string? jar-url) (string? pom-url)]}
   (println pom-url repos)
   (let [pom (pom/parse (slurp pom-url))
         project (util/clojars-id (pom/artifact-info pom))
--- a/modules/analysis-runner/src/cljdoc/analysis/deps.clj
+++ b/modules/analysis-runner/src/cljdoc/analysis/deps.clj
@@ -88,8 +88,7 @@
         (merge (get hardcoded-deps project))
         (ensure-required-deps)
         (ensure-recent-ish)
-        (merge cljdoc-codox)
-        (assoc project {:mvn/version version}))))
+        (merge cljdoc-codox))))
 
 (def ^:private default-repos
   {"central" {:url "https://repo1.maven.org/maven2/"},
@@ -103,15 +102,18 @@
 
 (defn resolved-and-cp
   "Build a classpath for the project specified by `pom-url`."
-  [pom-url repos]
+  [jar-url pom-url repos]
   {:pre [(string? pom-url)]}
   (println pom-url repos)
   (let [pom (pom/parse (slurp pom-url))
+        {:keys [group-id artifact-id version]} (pom/artifact-info pom)
+        project (symbol group-id artifact-id)
         resolved (tdeps/resolve-deps {:deps (deps pom),
                                       :mvn/repos (merge default-repos
                                                         (extra-repos pom)
                                                         repos)}
-                                     {:verbose false})]
+                                     {:extra-deps {project {:local/root jar-url}}
+                                      :verbose false})]
     {:resolved-deps resolved
      :classpath (tdeps/make-classpath resolved [] nil)}))
 
--- a/modules/analysis-runner/src/cljdoc/analysis/runner.clj
+++ b/modules/analysis-runner/src/cljdoc/analysis/runner.clj
@@ -156,7 +156,7 @@
   "Analyze the provided project"
   [{:keys [project version jarpath pompath repos]}]
   {:pre [(seq project) (seq version) (seq jarpath) (seq pompath)]}
-  (let [project (symbol project)
+  (let [project      (symbol project)
         analysis-dir (util/system-temp-dir (str "cljdoc-" project "-" version))
         output-file  (io/file util/analysis-output-prefix (util/cljdoc-edn project version))]
     (try
--- a/modules/analysis-runner/src/cljdoc/analysis/runner.clj
+++ b/modules/analysis-runner/src/cljdoc/analysis/runner.clj
@@ -87,16 +87,16 @@
                      jar)
         {:keys [classpath resolved-deps]} (deps/resolved-and-cp jar-path pom extra-mvn-repos)
         classpath* (str (.getAbsolutePath impl-src-dir) ":" classpath)]
-    (println "Used dependencies for analysis:")
-    (deps/print-tree resolved-deps)
-    (println "---------------------------------------------------------------------------")
     (unzip! jar-path jar-contents)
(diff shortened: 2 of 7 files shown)


residentsummer:
* fixed analysis for remote jars (broken in first commits of this PR)
* split jar preparation out of analysis-impl
* added test for local jars

modules/analysis-runner/src/cljdoc/analysis/runner_ng.clj
```diff
@@ -9,26 +9,21 @@
             [clojure.edn :as edn]
             [cljdoc.util :as util]
             [cljdoc.analysis.deps :as deps]
-            [cljdoc.analysis.runner :as runner]
+            [cljdoc.analysis.runner :as runner :refer [copy]]
```
martinklepsch: Not a big fan of using `:refer` in general because it hides the origin of a function. Explicit `:refer` is much better than `:refer :all` in that regard but IMHO it's best to just avoid `:refer` whenever possible.

modules/analysis-runner/src/cljdoc/analysis/runner.clj
```diff
@@ -20,8 +20,9 @@
            (java.net URI)
            (java.nio.file Files)))
 
-(defn copy [uri file]
-  (with-open [in  (io/input-stream uri)
+(defn copy [source file]
```
martinklepsch: Maybe this would be a reasonable thing to put into `cljdoc.util` since it's used in multiple places now.

modules/analysis-runner/src/cljdoc/analysis/deps.clj
```diff
@@ -103,15 +102,18 @@
 
 (defn resolved-and-cp
   "Build a classpath for the project specified by `pom-url`."
-  [pom-url repos]
+  [jar-url pom-url repos]
   {:pre [(string? pom-url)]}
   (println pom-url repos)
   (let [pom (pom/parse (slurp pom-url))
+        {:keys [group-id artifact-id version]} (pom/artifact-info pom)
+        project (symbol group-id artifact-id)
```
martinklepsch: You could use `cljdoc.util/clojars-id` here but it doesn't really matter much.

martinklepsch:
```clj
(util/clojars-id (pom/artifact-info pom))
```
Would give you a symbol that you can use in the deps map.

modules/analysis-runner/src/cljdoc/analysis/runner.clj
```diff
@@ -86,6 +77,34 @@
       (println "Deleting" path)
       (.delete file))))
 
+(defn- prepare-jar-for-analysis!
```
martinklepsch: I like how you pulled all that io stuff out of `analyze-impl` but I'm not sure if I think that the jar preparation work and class path calculation should be combined like this. An alternative idea: 

- `prepare-jar-for-analysis!` returns `{:jar-contents some-dir :jar-path local-jar-path}`
- `resolved-and-cp` is still called very early in the process (similar place like before)
- this would require `prepare-jar-for-analysis!` to also be called earlier

I think right now we have a call stack that is 

```
analyze-impl -> prepare-jar-for-analysis! -> resolved-and-cp
```
For testing in a REPL and also readability it could be nicer to have something more along the lines of this
```clj
(let [{:keys [jar-contents jar-path]} (prepare-jar-for-analysis! ,,,)
      {:keys [classpath]} (deps/resolved-and-cp jar-path ,,,)]
  (analyze-impl jar-contents classpath ,,,))
```

Let me know what you think :)

src/cljdoc/server/ingest.clj
```diff
@@ -14,10 +14,11 @@
 (defn ingest-cljdoc-edn
   "Store all the API-related information in the passed `cljdoc-edn` data"
   [storage {:keys [codox group-id artifact-id version] :as cljdoc-edn}]
-  (let [artifact (pom/artifact-info (pom/parse (:pom-str cljdoc-edn)))]
+  (let [project (util/clojars-id cljdoc-edn)
+        artifact (util/version-entity project version)]
```
martinklepsch: I think this change makes sense but we should also document it in the `cljdoc.cli` docs. Something like "The version you specify via the CLI will take precedence over the version specified in the POM file".

modules/analysis-runner/src/cljdoc/analysis/runner_ng.clj
residentsummer: resolved via moving `copy` to `util`

modules/analysis-runner/src/cljdoc/analysis/runner.clj
residentsummer: done

modules/analysis-runner/src/cljdoc/analysis/deps.clj
residentsummer: Thanks for the tip!

src/cljdoc/server/ingest.clj
residentsummer: Done. Please check if changes I've made to `ingest` command doc is ok :)

modules/analysis-runner/src/cljdoc/analysis/runner.clj
residentsummer: Hi, Martin!

Now I see that chaining functions like this will impede debugging in REPL. Sure, it's possible to rearrange calls to remove coupling and see the bigger picture. I got more questions, though :)

* I see that downloading and unpacking project JAR over and over again in `analyze-impl` (via `prepare-jar...`) will slow down debugging flow and obstruct the view on the process. But what are benefits of calculating CP separately from jar preparation?
* One point that bugged me in my current design is that `prepare-...` and `analyze-impl` are tightly coupled (`prepare` copies analysis files that will make sense only for `analyze-impl`). That's one of the reasons why call graph looks like this (the other reason is removing duplicate code from `-main` of `runner` and runner-ng`). Do you think it will be better to put copying those files back to `analysis-impl`?
* I got some hard time with temp dir management in this PR... Moving `prepare-jar-...` out of `analyze-impl`, will require some kind of conveying a shared temp dir to both of them. I've already done this for `prepare`, but it still feels kinda clumsy...

modules/analysis-runner/src/cljdoc/analysis/deps.clj
```diff
@@ -103,15 +102,17 @@
 
 (defn resolved-and-cp
   "Build a classpath for the project specified by `pom-url`."
-  [pom-url repos]
+  [jar-url pom-url repos]
   {:pre [(string? pom-url)]}
```
martinklepsch: A similar assertion could be added for the `jar-url` I guess.

modules/analysis-runner/src/cljdoc/analysis/runner.clj
martinklepsch: > But what are benefits of calculating CP separately from jar preparation?

Sometimes when debugging dependency resolution issues I download `.pom` files from Clojars and point the existing code at those files. I imagine something similar might come in handy once this lands. I don't have strong feelings about repeated downloading etc.

> One point that bugged me in my current design is that `prepare-...` and `analyze-impl` are tightly coupled

I'm not sure if there's a solution to this but maybe that's fine? After all the functions depend on another to get the job done.

> I got some hard time with temp dir management in this PR...

I actually missed that fact during my review and that of course makes things a little more complicated. The tmp-dir creation could be moved outside the `analyze-impl` `prepare-` functions as well but then again this adds more arguments to those functions. Could be nice to pass a local directory though, maybe a bit easier to find than system temp dirs.

martinklepsch:
Honestly I'm perfectly fine with the way it is and just saw some potential to make things a bit more apparent for people who look at this code for the first time. At this point there's no requested changes anymore so feel free to do as much or as little as you wish and let me know when you're done and we'll get things merged :)

modules/analysis-runner/src/cljdoc/analysis/deps.clj
residentsummer: done

residentsummer:
I've moved the call to `prepare-jar...` out of `analyze-impl` and packed the whole analysis flow into it's own function. Also removed redundant copying of analysis result file. There are changes for repl-based debugging as well - exposed some internal paths in the result of `prepare-jar-...` to make it possible to construct customized CP if needed.

> let me know when you're done and we'll get things merged :)
Martin, I think I'm done with it :)

martinklepsch:
Thanks @residentsummer! 🙌

PR #279, opened 2019-02-27
Only show code-view scroll bar when necessary

mhuebert:


martinklepsch:
Thanks a bunch Matt! I haven't noticed this myself but the change looks reasonable so just went ahead and merged it 🙂

PR #284, opened 2019-03-11
UX(docstrings): show docstring platform in margin. Closes #280.

### LINKED ISSUE #280: Only show docstring label if more than one exists
It would be nice if we could elide the "ClojureScript Docstring" label (see below) when only one docstring exists. 

It looks to me like the `ignore-nil?`-related logic here: https://github.com/cljdoc/cljdoc/blob/master/src/cljdoc/platforms.clj#L25
could be copied to the implementation of `varies?` here: https://github.com/cljdoc/cljdoc/blob/master/src/cljdoc/platforms.clj#L47

image

mhuebert:
Moved docstring platform indication into the (widened) left margin.

Before:
image

After:
image

### DIFF
--- a/src/cljdoc/render/api.clj
+++ b/src/cljdoc/render/api.clj
@@ -55,8 +55,8 @@
   (if (platf/varies? mp :doc)
     (for [p (sort (platf/platforms mp))
           :when (platf/get-field mp :doc p)]
-      [:div
-       [:span.f7.ttu.gray.db.nb2 (get {"clj" "Clojure" "cljs" "ClojureScript"} p) " docstring"]
+      [:div.relative
+       [:div.f7.gray.absolute.w3.nl5.pr2.tr.pv1 p]
        (some-> (platf/get-field mp :doc p) (docstring->html render-wiki-link))])
     (some-> (platf/get-field mp :doc) (docstring->html render-wiki-link))))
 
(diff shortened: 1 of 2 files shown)


martinklepsch:
Thanks @mhuebert 🙌

martinklepsch:
One suggestion for improvement would be how things handled on mobile or smaller screens. Perhaps on mobile the clj/cljs annotations should become block level elements? Maybe at the end of the docstring in superscript? 😄 

Screenshot 2019-03-12 at 16 36 24

PS. I fixed the link styling in docstrings.

mhuebert:
hmm, good question. we could increase the margin just a bit (so pl5-ns pl4) and they can fit, and on a small width that doesn't seem to look bad to me. Superscript at the end might be hard to spot, but superscript at the left? below are a couple of variants.

image

martinklepsch:
If you're willing to work on this I'll happily leave that decision to you. I do like the superscript/margin solution but could also get a bit crammed. Don't know. 😄  (

martinklepsch:
Unfortunately spotted another issue on article pages: 

C1452DC1-A986-4A76-BDB3-07F64EDDD9FF

martinklepsch:
Screenshot 2019-03-16 at 12 32 52

Made some changes in 15f34db46e2a51fd9ba926a7cb4851a8ef0816c0 so that it now looks like above.

PR #285, opened 2019-03-23
Handle {:doc false} when rendering doc string

jorinvo:
I came across this broken page:
https://cljdoc.org/d/buddy/buddy-core/1.5.0/api/buddy.core.mac

Screen Shot 2019-03-23 at 14 54 00

After being able to reproduce the same issue locally, I discovered what was causing the error:

The function [here](https://github.com/funcool/buddy-core/blob/master/src/buddy/core/mac.clj#L70) has a strange definition with a doc string which is then overwritten with `{:doc false}`.

I think cljdoc should handle `{:doc false}` in general.

Not sure what is the best way to handle it but with the fix, the page renders and the missing doc string looks like this:

Screen Shot 2019-03-23 at 14 50 11

And raw mode:

Screen Shot 2019-03-23 at 14 50 21

### DIFF
--- a/src/cljdoc/render/api.clj
+++ b/src/cljdoc/render/api.clj
@@ -34,11 +34,13 @@
 (defn docstring->html [doc-str render-wiki-link]
   [:div
    [:div.lh-copy.markdown
-    (-> doc-str
-        (rich-text/markdown-to-html
-         {:escape-html? true
-          :render-wiki-link (comp render-wiki-link parse-wiki-link)})
-        hiccup/raw)]
+    ;; If someone sets `{:doc false}`, there will be no docstring
+    (when doc-str
+      (-> doc-str
+          (rich-text/markdown-to-html
+            {:escape-html? true
+             :render-wiki-link (comp render-wiki-link parse-wiki-link)})
+          hiccup/raw))]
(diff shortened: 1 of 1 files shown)


src/cljdoc/render/api.clj
```diff
@@ -34,11 +34,13 @@
...
   [:div
    [:div.lh-copy.markdown
-    (-> doc-str
-        (rich-text/markdown-to-html
-         {:escape-html? true
-          :render-wiki-link (comp render-wiki-link parse-wiki-link)})
-        hiccup/raw)]
+    (if (= doc-str false)
+      ""
+      (-> doc-str
```
martinklepsch: You could also just do
```clj
(when doc-str
  (-> doc-str ,,,))
```
In hiccup an empty string and nil are treated identically in most cases I think.

martinklepsch:
Oh and :clap: for also opening a PR to the actual project!

src/cljdoc/render/api.clj
jorinvo: Wasn't sure if `nil` works as result here. Changed now and result still looks the same :)

jorinvo:
I followed your suggestion to simplify the code and also added a small comment. Merging now. Thanks for your fast feedback @martinklepsch :)

martinklepsch:
Happy to! Deployed & works 👍

PR #286, opened 2019-03-26
Integrate build status into badge + docs pages

### LINKED ISSUE #243: Make badge indicate that there is a problem if last build failed
Sometimes builds fail and it would be useful to indicate that within the cljdoc badge. Could look something like this: 

```
![](https://badgen.now.sh/badge/cljdoc/1.1.1-problem/red)
```

Since the URL will still link to the regular documentation page some link to the most recent build (if it failed) needs to be integrated into the documentation pages. Maybe similar to the link to more recent versions:

screenshot 2018-11-28 at 10 11 59

### Pointers

- Badge rendering code:
  https://github.com/cljdoc/cljdoc/blob/139a5a6c8094ef05cde24b93ef942b0882dbb22f/src/cljdoc/server/pedestal.clj#L206-L208

  - This will require the `build-tracker` system component to retrieve the latest build for the requested version.

- To include information about the last build we could add a new interceptor that adds the last build to the context. 
  - Currently most rendering code only gets the `cache-bundle` but I started to [fade that out](https://github.com/cljdoc/cljdoc/commit/daa811db923c24d31b2644a7f01d587dd994051f) and it might make sense to loosen this up further.
  - Here's an example of an interceptor loading some data into the context:
       https://github.com/cljdoc/cljdoc/blob/139a5a6c8094ef05cde24b93ef942b0882dbb22f/src/cljdoc/server/pedestal.clj#L95-L106

martinklepsch:
The badges are a useful way to provide information about the state of a project to it's authors and other users. To make it even more useful it should display information about the build status if there has been a problem.

- [x] Show if API/Git import failed in badge message
- [x] Integrate warning into docs pages if there were problems in last build 

Fixes #243 

#### Demo screenshots (docs page + badge):

Screenshot 2019-03-27 at 11 52 29
Screenshot 2019-03-27 at 11 54 07

Screenshot 2019-03-27 at 11 52 41
Screenshot 2019-03-27 at 11 53 48

### DIFF
--- a/src/cljdoc/server/build_log.clj
+++ b/src/cljdoc/server/build_log.clj
@@ -20,7 +20,8 @@
   (completed! [_ build-id])
   (get-build [_ build-id])
   (recent-builds [_ days])
-  (running-build [_ group-id artifact-id version]))
+  (running-build [_ group-id artifact-id version])
+  (last-build [_ group-id artifact-id version]))
 
 (defrecord SQLBuildTracker [db-spec]
   IBuildTracker
@@ -87,7 +88,28 @@
                               "order by id desc "
                               "limit 1")
                          group-id artifact-id version
-                         (str (.minus (Instant/now) (Duration/ofMinutes 10)))]))))
+                         (str (.minus (Instant/now) (Duration/ofMinutes 10)))])))
+  (last-build [_ group-id artifact-id version]
+    (first
+     (sql/query db-spec [(str "select * from builds where "
+                              "(group_id is ? "
+                              "and artifact_id = ? "
+                              "and version = ?) "
+                              "and (import_completed_ts is not null "
+                              "or error is not null) "
+                              "order by id desc "
+                              "limit 1")
+                         group-id artifact-id version]))))
+
+
+(defn api-import-successful? [build]
+  (and (:api_imported_ts build)
+       (nil? (:error build))))
+
+(defn git-import-successful? [build]
+  (and (:scm_url build)
+       (:git_imported_ts build)
+       (nil? (:git_problem build))))
 
 (comment
   (require 'ragtime.repl 'ragtime.jdbc)
--- a/src/cljdoc/server/pedestal.clj
+++ b/src/cljdoc/server/pedestal.clj
@@ -62,7 +62,9 @@
                   (assoc ctx :response {:status 302, :headers {"Location" location}}))
 
                 (if cache-bundle
-                  (pu/ok-html ctx (html/render page-type path-params cache-bundle))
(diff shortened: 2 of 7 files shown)


src/cljdoc/server/build_log.clj
```diff
@@ -87,7 +88,28 @@
                               "order by id desc "
                               "limit 1")
                          group-id artifact-id version
-                         (str (.minus (Instant/now) (Duration/ofMinutes 10)))]))))
+                         (str (.minus (Instant/now) (Duration/ofMinutes 10)))])))
+  (last-build [_ group-id artifact-id version]
+    (first
```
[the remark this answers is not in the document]
dawran6: To me `first` is clear enough, but I don't have a strong opinion on this.

src/cljdoc/server/pedestal.clj
```diff
@@ -226,28 +235,35 @@
...
+    (->> {:status 200
+          :headers {"Content-Type" "image/svg+xml;charset=utf-8"
+                    "Cache-Control" (format "public; max-age=%s" (* 30 60))}
+          :body (:body (clj-http.lite.client/get url {:headers {"User-Agent" "clj-http-lite"}}))}
+         (assoc ctx :response))))
 
 (defn badge-interceptor []
   {:name ::badge
-   :enter (fn badge [ctx]
+   :leave (fn badge [ctx]
```
dawran6: May I know what's the rationale behind implementing `:leave` instead of `:enter` for this interceptor? I just find keeping all my interceptors as the `enter`-kind easier to understand. I'd love to know any good use case for the `leave`-kind interceptors.

martinklepsch: Totally, great question. 👍 This interceptor also has an `:error` clause now to catch any exceptions from the resolve-version-interceptor, rendering a badge that shows a proper error message.

```clj
         :badge-for-project  [(badge-interceptor)
	                      resolve-version-interceptor
	                      (last-build-loader build-tracker)])
```
If we kept it as an `:enter` interceptor it would only be queued after the `resolve-version-interceptor` meaning the `:error` handler would also not take effect because the interceptor hasn't been ran whenever `resolve-version-interceptor` would throw an exception.

In the other places where `resolve-version-interceptor` is used the exception throwing is will result in a 500, which is the current behaviour.

Let me know if this makes sense 😄

martinklepsch: FWIW I'm also not super happy about mixing `:enter` and `:leave` too much but it's the best I could come up with that still works with the other `resolve-version-interceptor` use cases.

dawran6: Thanks, it totally makes sense. :)

PR #293, opened 2019-04-29
Render wikilinks as code. Issue #288.

### LINKED ISSUE #288: Render wikilinks as code
Currently wikilinks are rendered as regular links but since they always refer to other vars we should also render them as `<code>`.

An example: https://cljdoc.org/d/rum/rum/0.11.3/api/rum.core#cursor

Screenshot 2019-04-11 at 10 24 50

I think this would be possible by either adding a class to those links right in the LinkResolver code as described [in this discussion](https://gitq.com/vsch/flexmark-java/topics/10/hooks-into-wikilinks-extension-rendering/4) or by customising the rendering of wikilinks. I think the letter is described [here](https://github.com/vsch/flexmark-java/issues/140) but not 100% sure. 

I have a slight preference for the latter approach but would be happy to merge a PR doing either. 

In the end it should look something like this:

Screenshot 2019-04-11 at 10 28 21

anthonygalea:
If I understood correctly the goal here is to wrap the anchor text in a `<code>` element. Here's a screenshot showing before and after:
Screenshot 2019-04-29 at 21 08 12

I used the following as guides:
* [PegdownCustomLinkResolverOptions](https://github.com/vsch/flexmark-java/blob/master/flexmark-java-samples/src/com/vladsch/flexmark/samples/PegdownCustomLinkResolverOptions.java)
* [WikiLinkNodeRenderer](https://github.com/vsch/flexmark-java/blob/master/flexmark-ext-wikilink/src/main/java/com/vladsch/flexmark/ext/wikilink/internal/WikiLinkNodeRenderer.java#L44)

fixes #288


martinklepsch:
Cherry picked this into master, thanks! 🎉 

Send me your address via Slack if you want some stickers! 🙂 

(PS. I really hate the fact that the contribution was accepted but because I rebased and merged locally this is not properly reflected on GitHub 🙁)

PR #296, opened 2019-05-20
feature documentation contributors at the end of articles

martinklepsch:
Inspired by [Daniel's tweet](https://twitter.com/danielwithmusic/status/1130559802680250372) and a two hour train ride with no internet. 

Screenshot 2019-05-21 at 00 13 35

> 🐛 **Note:** Since this is something that is analysed and stored at build time it will only be available for builds done after this has been deployed.


martinklepsch:
Merged this for now but if anyone has thoughts on the diff, please don't hold them back :)

Issue #236, opened 2018-11-26
How to generate previews for PRs using CI systems?


urzds:
Situation: Someone submits a PR to my project (which has its documentation on cljdoc) to improve documentation. I want to preview this documentation to make sure it contains no formatting issues and renders as intended. I already have a CI system set up for my project that can build the source code and run tests.

Question: How do I generate a preview of how the changes will look like on cljdoc? Both I and the contributor should be able to look at it (i.e. the result should be available on some web space), to discuss further improvements or catch formatting mistakes.

Thoughts:
* https://github.com/cljdoc/cljdoc/blob/master/doc/running-cljdoc-locally.md
  - The process seems involved and appears to depend on a continuously running cljdoc server.
* Envisioned process:
  1. A command line tool that can generate static HTML pages
  2. Which I can then store as a build artifact (i.e. it persists after the build job completed)
  3. And push to a simple web server
  4. Where me and the contributor can have a look at it

martinklepsch:
A nice way to integrate this into people's CI workflow could be by exposing it as a [CircleCI Orb](https://circleci.com/docs/2.0/creating-orbs/). Obviously this will only work for CircleCI :smile:

martinklepsch:
I've looked into this a bit more just now. One of the things that's still unclear to me is how/if the orb would have access to the workspace/disk as it was left by previous CI steps. cljdoc would need to read the jar file, ideally just have access to the entire `~/.m2`. 

https://circleci.com/docs/2.0/creating-orbs/

PR #301, opened 2019-05-24
Documentation about running locally with Docker

borkdude:
@martinklepsch Please test the updated instructions before merging :-)

One thing to keep in mind: the docs assume that there is a `latest` tag on Dockerhub which points to the latest image (this is common). I've updated the `Makefile` to create the `latest` tag and I'm assuming that the `.circleci/config.yml` will push that tag to Dockerhub.

### DIFF
--- a/doc/running-cljdoc-locally.md
+++ b/doc/running-cljdoc-locally.md
@@ -98,13 +98,27 @@ for `muuntaja` from a local Jar and Git repository:
 
 1. Open the docs for muuntaja on the local cljdoc server: http://localhost:8000/d/metosin/muuntaja
 
-# Running with Docker
+# Running locally with Docker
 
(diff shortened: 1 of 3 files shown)


borkdude:
@martinklepsch Please test the updated instructions before merging :-D

martinklepsch:
Thanks a lot Michiel, this looks great!

There's been work on something (slightly) related, which is exposing cljdoc as a CircleCI orb so people can build docs during CI and view them as part of their CI pipeline: https://github.com/cljdoc/cljdoc/issues/236. It seems that would cover the use case you have with clj-kondo so if it's something you'd love to have, feel free to contribute in that area 🙂 

@all-contributors add @borkdude for docs.

PR #311, opened 2019-06-06
Search: fix err when no origin

holyjak:
Fix https://sentry.io/share/issue/afa0d0b8e99d44d3a1b100f3fd3dd2b6/

> llegalArgumentException: value must not be null

at artifact->doc line 126

### DIFF
--- a/src/cljdoc/server/search/artifact_indexer.clj
+++ b/src/cljdoc/server/search/artifact_indexer.clj
@@ -116,14 +116,15 @@
   [{:keys [^String artifact-id
            ^String group-id
            ^String description
+           ^String origin
            versions]
-    :or {description ""}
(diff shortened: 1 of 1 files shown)


src/cljdoc/server/search/artifact_indexer.clj
```diff
@@ -116,14 +116,15 @@
   [{:keys [^String artifact-id
            ^String group-id
            ^String description
+           ^String origin
            versions]
-    :or {description ""}
+    :or {description "", origin "N/A"}
```
martinklepsch: Aren't we always setting `origin` to "Clojars" or "Maven Central" though?

holyjak:
Not when you call `(index-artifact..) `. We could / should also change the
release monitor to supply it.

PR #316, opened 2019-06-13
Search: suggestions API for OpenSearch

holyjak:
Inspired by PR #224 to provide suggestions for OpenSearch, e.g. in
Firefox, see https://developer.mozilla.org/en-US/docs/Web/OpenSearch

### DIFF
--- a/src/cljdoc/server/pedestal.clj
+++ b/src/cljdoc/server/pedestal.clj
@@ -16,7 +16,7 @@
             [cljdoc.render.build-log :as render-build-log]
             [cljdoc.render.index-pages :as index-pages]
             [cljdoc.render.home :as render-home]
-            [cljdoc.render.search :as search]
+            [cljdoc.render.search :as render-search]
             [cljdoc.render.meta :as render-meta]
             [cljdoc.render.error :as error]
             [cljdoc.render.offline :as offline]
@@ -26,7 +26,7 @@
             [cljdoc.server.pedestal-util :as pu]
             [cljdoc.server.routes :as routes]
             [cljdoc.server.api :as api]
-            [cljdoc.server.search.api :as sc]
+            [cljdoc.server.search.api :as search-api]
             [cljdoc.server.sitemap :as sitemap]
             [cljdoc.server.ingest :as ingest]
             [cljdoc.storage.api :as storage]
@@ -222,7 +222,16 @@
             (if-let [q (-> ctx :request :params :q)]
               (assoc ctx :response {:status  200
                                     :headers {"Content-Type" "application/json"}
-                                    :body    (sc/search searcher q)})
+                                    :body    (search-api/search searcher q)})
+              (assoc ctx :response {:status 400 :headers {} :body "ERROR: Missing q query param"})))})
+
+(defn search-suggest-interceptor [searcher]
+  {:name  ::search-suggest
+   :enter (fn search-suggest-handler [ctx]
+            (if-let [q (-> ctx :request :params :q)]
+              (assoc ctx :response {:status  200
+                                    :headers {"Content-Type" "application/x-suggestions+json"}
+                                    :body    (search-api/suggest searcher q)})
               (assoc ctx :response {:status 400 :headers {} :body "ERROR: Missing q query param"})))})
 
 (defn show-build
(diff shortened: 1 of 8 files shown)


holyjak:
@martinklepsch I would appreciate your feedback on the integration into the rest of the system, especially the URL path.

src/cljdoc/server/routes.clj
```diff
@@ -41,8 +42,7 @@
     ["/versions/:group-id/:artifact-id" :get nop :route-name :artifact/index]})
 
 (defn open-search-routes []
-  #{["/search" :get nop :route-name :search]
-    ["/suggest" :get nop :route-name :suggest]})
+  #{["/search" :get nop :route-name :search]})
```
holyjak: Q: Rename `open-search-routes` to `search-pages-routes`?

src/cljdoc/server/pedestal.clj
```diff
@@ -372,8 +381,7 @@
    {:keys [route-name] :as route}]
   (->> (case route-name
          :home       [{:name ::home :enter #(pu/ok-html % (render-home/home))}]
-         :search     [{:name ::search :enter #(pu/ok-html % (search/search-page %))}]
-         :suggest    [{:name ::suggest :enter search/suggest-api}]
```
holyjak: I think the suggest API should be under /api because it returns JSON, not a page. Also, I think search-suggest is better then just suggest to make it clear it is in the context of search, though I am very open to discussion. Thoughts?

src/cljdoc/server/search/search.clj
```diff
@@ -134,19 +141,19 @@
...
-          ;parser           (QueryParser. "artifact-id" analyzer)
-          ;query            (.parse parser query-str)
-          ^Query query     (if (instance? Query query-in)
-                             query-in
-                             (parse-query query-in))
-          ^TopDocs topdocs (.search searcher query page-size)
-          hits             (.scoreDocs topdocs)]
-      (f {:topdocs topdocs :score-docs hits :searcher searcher :query query}))))
+(defn search->results
+  ([^String index-dir ^String query-in f] (search->results index-dir query-in 30 f))
```
martinklepsch:
```suggestion
  ([^String index-dir ^String query-in f] 
   (search->results index-dir query-in 30 f))
```
maybe?

src/cljdoc/server/search/search.clj
```diff
@@ -134,19 +141,19 @@
...
+   (with-open [reader (DirectoryReader/open (fsdir index-dir))]
+     (let [searcher         (IndexSearcher. reader)
+           ;parser           (QueryParser. "artifact-id" analyzer)
+           ;query            (.parse parser query-str)
+           ^Query query     (if (instance? Query query-in)
+                              query-in
+                              (parse-query query-in))
+           ^TopDocs topdocs (.search searcher query max-results)
+           hits             (.scoreDocs topdocs)]
+       (f {:topdocs topdocs :score-docs hits :searcher searcher :query query})))))
```
martinklepsch: This is slightly unrelated to the change of this PR but passing in `f` here seems odd to me. The difference in usage seems to be

```clj
(search->results ,,, f)
;; vs
(f (search->results ,,,))
```
I find the latter much clearer in terms of understanding how `f` is used. Passing it in makes that a little less obvious. My 2 cents anyway, feel free to ignore :)

src/cljdoc/server/system.clj
```diff
@@ -114,6 +114,15 @@
   ;; This is the main REPL entry point into cljdoc.
   ;; Run the forms one by one up to `go` and you should
   ;; have a running server at http://localhost:8000
+  ;; OR run then all at once:
+  (do
+    (require '[cljdoc.server.system])
+    (in-ns 'cljdoc.server.system)
+    (require '[integrant.repl])
+    (integrant.repl/set-prep! #(system-config (cfg/config)))
+    (integrant.repl/go))
```
martinklepsch: This is better, feel free to just delete the duplicate lines below this.

src/cljdoc/server/routes.clj
martinklepsch: Sounds good 👍

src/cljdoc/server/pedestal.clj
martinklepsch: Fair point. I think arguments could be made for either location but if you feel that it makes more sense under `/api/` I don't have anything to object.

src/cljdoc/server/search/search.clj
holyjak: I have added a clarification comment. It needs to be that way because it must be run within the `search`'s `with-open`, it cannot run after the search finished.

martinklepsch: Ah interesting, didn’t consider this :)

holyjak: Fun fact: me too :) Coming back to the code to implement the suggest API, I looked at `search` and had the exact same thought as you. After factoring `f` out and calling it, I got the "index reader already closed" exception which reminded me of why did it this way in the first place. A good reminder that any code deviating from what is normal needs to be well explained :)

PR #317, opened 2019-06-14
Show stacktrace for import errors; fixes #111

### LINKED ISSUE #111: Show exceptions for failing builds
Builds may fail for two reasons:

1. The Clojure analysis not working (e.g. missing dependency)
2. The Git analysis not working (e.g. missing file that is listed in doctree)

The first happens in the Circle CI build job, the second on the cljdoc server.

Storing the exception for the 2nd case is relatively easy as they are already within our Clojure runtime. For CircleCI we'd have to get the logs etc. Probably just addressing case 2 is a good enough start because CircleCI provides it's own interface to view logs.

jacobobryant:
Example:
image

Happy to make changes if needed, just let me know.

### DIFF
--- a/src/cljdoc/render/build_log.clj
+++ b/src/cljdoc/render/build_log.clj
@@ -171,6 +171,8 @@
            ""
            [:h3.mt0 "There was an error"]
            [:p.bg-washed-red.pa3.br2 (:error build-info)]
+           (when (:error_info build-info)
+             [:pre.overflow-x-scroll.f6 (:error_info build-info)])
            (when (:analysis_job_uri build-info)
              [:p.lh-copy "Please see the " [:a.link.blue {:href (:analysis_job_uri build-info)}
               "build job"] " to understand why this build failed and reach out if you aren't
--- a/src/cljdoc/render/build_log.clj
+++ b/src/cljdoc/render/build_log.clj
@@ -3,7 +3,9 @@
             [cljdoc.render.layout :as layout]
             [cljdoc.util :as util]
             [cljdoc.util.datetime :as dt]
-            [cljdoc.server.routes :as routes])
+            [cljdoc.server.routes :as routes]
+            [taoensso.nippy :as nippy]
+            [clojure.stacktrace :refer [print-stack-trace]])
   (:import [java.time Instant Duration ZoneId]
            [java.time.temporal ChronoUnit]))
 
@@ -172,7 +174,10 @@
            [:h3.mt0 "There was an error"]
            [:p.bg-washed-red.pa3.br2 (:error build-info)]
            (when (:error_info build-info)
-             [:pre.overflow-x-scroll.f6 (:error_info build-info)])
+             [:pre.f6 [:code (with-out-str (-> build-info
+                                               :error_info
+                                               nippy/thaw
+                                               print-stack-trace))]])
            (when (:analysis_job_uri build-info)
              [:p.lh-copy "Please see the " [:a.link.blue {:href (:analysis_job_uri build-info)}
               "build job"] " to understand why this build failed and reach out if you aren't
--- a/src/cljdoc/render/build_log.clj
+++ b/src/cljdoc/render/build_log.clj
@@ -1,11 +1,12 @@
 (ns cljdoc.render.build-log
-  (:require [clojure.string :as string]
+  (:require [clojure.stacktrace :as stacktrace]
+            [clojure.string :as string]
             [cljdoc.render.layout :as layout]
             [cljdoc.util :as util]
             [cljdoc.util.datetime :as dt]
             [cljdoc.server.routes :as routes]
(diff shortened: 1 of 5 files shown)


martinklepsch:
This looks fantastic @jacobobyrant! 

I’d have a few ideas to improve it further:

- use a code block to render the stack trace 
- Format the stacktrace to its string representation, something along the lines of `.printStackTrace` 
- (maybe) incorporate `ex-data` information into the UI somehow (if that’s possible at all) 
- potentially serialize the exception using nippy to be able to call `ex-data` on it etc. 

Super excited to see this come together, this will make it much easier for people to understand why their build failed and open appropriate issues 👌🙌

jacobobryant:
:+1: @martinklepsch I used nippy + `print-stack-trace` + `code`, but not sure about `ex-data`. Do you mean wrap the exception with more contextual information before serializing it, e.g. `(ex-info "Exception during import" {:project "foo" :version "bar"} e)`? If so, what information would you want to put in there?

Also, a small note--the serializing & deserializing is a little unbalanced. I serialize the exception inside `SQLBuildTracker` since there's a single point of entry (`failed!`), but I deserialize outside of `SQLBuildTracker` because otherwise we'd have to either:
1. update all of the query fns
2. update just `get-build`, leaving the other query functions with an inconsistent return value (i.e. `get-build` returns `:error_info` deserialized but the other fns return it still serialized).

So I opted to simply have `SQLBuildTracker` always return `:error_info` still serialized and let the caller have responsibility for deserializing. Are you fine with this, or should I deserialize within `SQLBuildTracker`?

image

src/cljdoc/render/build_log.clj
```diff
@@ -3,7 +3,9 @@
             [cljdoc.render.layout :as layout]
             [cljdoc.util :as util]
             [cljdoc.util.datetime :as dt]
-            [cljdoc.server.routes :as routes])
+            [cljdoc.server.routes :as routes]
+            [taoensso.nippy :as nippy]
+            [clojure.stacktrace :refer [print-stack-trace]])
```
martinklepsch: I'd suggest to only use `:refer` for very commonly used symbols as described here: https://stuartsierra.com/2016/clojure-how-to-ns.html

src/cljdoc/render/build_log.clj
```diff
@@ -171,6 +173,11 @@
            ""
            [:h3.mt0 "There was an error"]
            [:p.bg-washed-red.pa3.br2 (:error build-info)]
+           (when (:error_info build-info)
+             [:pre.f6 [:code (with-out-str (-> build-info
```
martinklepsch: Consider this set of classes to get the typical "code block" style look that's used across the site: `:pre.lh-copy.bg-near-white.code.pa3.br2.f6.overflow-x-scroll`.

martinklepsch:
With `ex-data` I meant that if an exception has additional structured data (as returned by `ex-data`) we should somehow make that visible to the user viewing the page. In many cases this may not be the case but in some cases this information might be vital to diagnosing the issue. Having this in place may also allow cljdoc to surface troubleshooting information in a more structured way in the future.

I'm imagining a separate code block with the `ex-data` pretty printed (using zprint, see [`api.clj`](https://github.com/cljdoc/cljdoc/blob/master/src/cljdoc/render/api.clj#L21)) but other approaches would also be fine. 

It could also make sense to add a `.nohighlight` class to the stack trace container to prevent highlight.js from trying to highlight it as code.

The decisions you made around de/serializing are fine with me. Eventually we might want to consolidate this but for now this is good 👍

jacobobryant:
Here's the latest iteration. Any preferences on formatting for ex-data?

image

martinklepsch:
This looks great! I don't have any immediate suggestions on how to improve the formatting of the ex-data further and think this gets the job done :) 

I've added you to the cljdoc community on GitHub, which means you have write access to the repository now. Please merge your PR whenever you feel it is ready. It will then be deployed automatically via CI. Welcome to the gang! 😊

martinklepsch:
Oh and if you're wondering whether to squash or not: I don't have any strong opinions on this, the rules I follow for myself are:

- If there's multiple commits, rebase and merge with merge commit
- If there's only one commit then squash & merge

Generally I don't mind Git history being a bit more verbose as long as the main branch is linear (i.e. things are rebased before merging).

jacobobryant:
Sounds good, and thanks!

martinklepsch:
Hey! Unfortunately I just noticed that we ended up having two migrations with the `008` index. If you'd be willing to help fix this, that would be fantastic. Suggested actions:

- rename the migration added as part of this PR to `009-add-error-info-column-to-builds`
- (optional) ensure stuff like this breaks the build

jacobobryant:
Sorry about that; I'm not sure how I missed the existing `008-*` migration. I see you've renamed the migration already. I'll create a PR with a test for this.

Issue #327, opened 2019-07-15
Are we happy with asciidoc admonitions formatting?


lread:
**Current behavior**
Given the following asciidoc:
```asciidoc
== Admonitions

NOTE: Note that this note is a note.

TIP: Tip the scales with a tip.

IMPORTANT: Important things are said here.

CAUTION: Aren't you a caution?

WARNING: Warning to the wise.
```
cljdoc currently produces:
![image](https://user-images.githubusercontent.com/967328/61221564-c936a000-a6e6-11e9-83cf-e8efe6b7524b.png)

**Behavior to consider**
Let's compare this to the asciidoc browser extension rendering (which matches asciidoctor user guide renderings):
![image](https://user-images.githubusercontent.com/967328/61221576-d3589e80-a6e6-11e9-975f-54311198caec.png)

I don't know if going with text and coloring over icons was a conscious design decision so I bring it up in this issue for discussion.

martinklepsch:
This was a partly conscious decision and the same applies to #321 as you already pointed out.

I find Asciidoctor's standard styles could be more prominent. I'm also not sure if/how the icons will work with offline documentation. Are they loaded from a server or embedded into the document somehow? 

Generally matching Asciidoctor's output exactly hasn't been a goal for me but it is important that most Asciidoc features can be used and are supported as in that they can be used by authors to guide readers.

lread:
It looks like [offline viewing is supported](https://asciidoctor.org/docs/user-manual/#icons)

> If you want icons to display using images, set the icons attribute to an empty value in the document header. This strategy is recommended if you are converting to DocBook or you want an easy way to make HTML for viewing offline. The DocBook toolchain provides images for the admonition and callout icons, which you can replace with your own custom icons. If you use the inline icon macro, you’ll need to provide the images for those icons.

But we'd have to test, of course.

lread:
I think I have found a way to stick with the textual representation while still supporting #321, and I propose, unless there is strong interest in icons, that we do that for now.

I suppose we should style the currently unstyled CAUTION admonition (see screenshot above).

If I look at the spirit of asciidoc admonitions, it seems a bit fuzzy but maybe they were inspired by ansi safety standards [which I found summarized here](https://www.mysafetysign.com/ansi-safety-sign-standards). Here's my thinking on coloring:
1. IMPORTANT (might be = Danger?) - death possible RED (currently yellow)
1. WARNING - serious injury possible ORANGE (currently red)
[...]

Issue #324, opened 2019-07-15
Asciidoc tables need some formatting love.


lread:
**Current behavior**
Given the following asciidoc 
```asciidoc
Asciidoc tables are quite full featured, here's a simple one:

|===
| Heading 1 | Heading 2

| col1, row1
| col2, row1

| col1, row2
| col2, row2

| col1, row3
| col2, row3

| col1, row4
| col2, row4
|===

Here's the same table with a title

.This time with a title
|===
| Heading 1 | Heading 2

| col1, row1
| col2, row1

| col1, row2
| col2, row2
|===

And here's a nested table example from asciidoctor user manual:

[cols="1,2a"]
|===
| Col 1 | Col 2

| Cell 1.1
| Cell 1.2

| Cell 2.1
| Cell 2.2

[cols="2,1"]
!===
! Col1 ! Col2

! C11
! C12

!===

|===
```

We currently get:
![image](https://user-images.githubusercontent.com/967328/61218495-2418c900-a6e0-11e9-96ec-2e8aa0618ce1.png)

**Desired behavior**

Let's compare this to the asciidoc browser extension rendering (which matches asciidoctor user guide renderings):
![image](https://user-images.githubusercontent.com/967328/61218582-5e826600-a6e0-11e9-8835-669f28722cf0.png)

And also compare to rendering of tables form CommonMark in a docstring:
![image](https://user-images.githubusercontent.com/967328/61218668-82de4280-a6e0-11e9-9e82-48dfbd928068.png)

A consistent (or at least similar) rendering of tables from CommonMark and asciidoc would be desirable.

martinklepsch:
Absolutely agree, wasn't aware that that was missing 😲

lread:
asciidoc tables are full featured and require quite a bit of css which I am bringing in from [asciidoctor default styles](https://github.com/asciidoctor/asciidoctor/blob/master/data/stylesheets/asciidoctor-default.css#L3).  

This is on the painstaking side and we will want to figure out how to streamline this.  

An issue is that the asciidoc style sheet not only brings in a general look and feel, but also supports asciidoc features that authors will use in their documents.

Perhaps bringing in the default asciidoc stylesheet, then programatically tweaking (and css scoping) it to become cljdocified would be a worthy experiment - unless I run into real trouble, I will leave this out of scope for this issue.

lread:
Here's some sample renderings:
[...]

Issue #323, opened 2019-07-15
Code block titles look like regular text


lread:
**Current behavior**
An asciidoc code block can have a title:
```asciidoc
.code-block-title
[source,clojure]
----
;; some sweet clojure code

(defn hello-earthling
  "Greet an earthling in a believable way."  ;; <1>
  [name]
  (println "Hello earthling " (uppercase name))) ;; <2>
----
<1> a docstring comes before the arguments
<2> check with an earthling if uppercasing is soothing
```
cljdoc currently renders this as:
![image](https://user-images.githubusercontent.com/967328/61218151-5970e700-a6df-11e9-9b24-cd05b98e0a42.png)

Which does not at all stand out from regular text.

**Desired behavior**
The asciidoc browser extension (and asciidoctor user guide) distinguish the title like so:
![image](https://user-images.githubusercontent.com/967328/61218237-9046fd00-a6df-11e9-8d45-3779ca204875.png)

Notice also that the code block title hugs the top of the code block.

martinklepsch:
👍

lread:
Here's what I've got so far, nothing drastic, just respecting that it is a title and conforming to our color scheme:
![image](https://user-images.githubusercontent.com/967328/61478611-d6aa8f00-a95f-11e9-9452-a1352a2383ef.png)

Issue #322, opened 2019-07-15
Give asciidoc table of contents title some oomph


lread:
**Current behavior**
An asciidoc can optionally automatically include a table of contents via `:toc:`.  cljdoc currently renders an asciidoc toc like so:
![image](https://user-images.githubusercontent.com/967328/61217342-75738900-a6dd-11e9-9f24-5b29ad446ca1.png)

I feel the toc title is not getting the formatting it deserves.

**Desired behavior**
Compare this to formatting from the asciidoc browser plugin (which matches the asciidoc user guide):
![image](https://user-images.githubusercontent.com/967328/61217900-ba4bef80-a6de-11e9-9965-255e912de5ec.png)

martinklepsch:
Yeah that would also be nice. Although we'll have to balance it with the TOC in the sidebar. But I guess we'll manage :)

lread:
Here's my current work in progress. Nothing drastic at all, just respecting that the toc title is a header:
![image](https://user-images.githubusercontent.com/967328/61478421-787dac00-a95f-11e9-8a3d-d072269fb8f6.png)

lread:
Ok not to bore you, but just in case anybody doth want to protest, my latest asciidoc toc styling looks like so:
![image](https://user-images.githubusercontent.com/967328/61919545-1fe07d00-af24-11e9-80a3-d879ecf37715.png)

PR #334, opened 2019-07-26
asciidoc formatting

### LINKED ISSUE #321: Render AsciiDoc code callouts
**Current behaviour**
cljdoc renders the following asciidoc callouts:

```asciidoc
[source,clojure]
----
;; some sweet clojure code

(defn hello-earthling
  "Greet an earthling in a believable way."  ;; <1>
  [name]
  (println "Hello earthling " (uppercase name))) ;; <2>
----
<1> a docstring comes before the arguments
<2> check with an earthling if uppercasing is soothing
```
like so:
image

As you can see the callouts rendered as (1) and (2) in the code do not catch the eye very well.

**Desired behavior**
The asciidoctor user manual shows callouts  rendered as icons, and so does the browser extension renderer:
image

The callouts rendered as icons are much easier to see.
(Ignore the fact that clojure code is not syntax highlighted here, cljdoc will continue to syntax highlight)

lread:
- fixed formatting for tables, admonitions, code callouts, content titles,
toc title
- upgraded highlightjs to current version
- now including highlighjs support for asciidoc as we are likely to include
asciidoc snippets in our user guides.
- asciidoc compatible v4.7.0 of font-awesome added
  - this was not required for our current admonitions or callouts but allows
  checkboxes and other icons to render properly.
  - decided on hosting this ourselves instead of going CDN route because CDN
  route seems to require account setup with font awesome?
- asciidoc css is now brought in by sass generated `cljdoc-asciidoc.css`
  - we use the same base sass that asciidoctor does for its default stylesheet
  which comes from https://github.com/asciidoctor/asciidoctor-stylesheet-factory
  - we take advantage of tachyons variable to encourage us to stick with
  tachyons stylings when we can
  - noticed some random styling issues in addition to ones reported in git
  issues and provided fixes - see `resources/public/sass/cljcdoc-asciidoc.scss`
  for details.
- offline support
  - includes new assets
  - now styles asciidoc documents (was not doing this before)
- see `resources/public/sass/README.md` for build instructions and further
details

Out of scope for this commit, but to consider:
- move remainder of styling in `cljdoc.css` to sass.  Benefits
  - specify tachyons styling elements by name
  - keeping css styling in one spot.
- include sass building in greater cljdoc build. It is currently a manually
invoked process.
- introduce auto-reloading of css and sass while developing ala figwheel main
or similar.
- include images in offline bundle

fixes #321 #322 #323 #324 #327


martinklepsch:
Thanks Lee! I think making the formatting more consistent is great but I'm a little worried about the increased complexity by adding SASS to the build steps. Is there a way we could achieve proper formatting and offline support without bringing this tooling into the project?

Also it seems there is a CDN for font-awesome: https://www.bootstrapcdn.com/fontawesome/ ? 

If you've squashed this commit from multiple commits, please consider maintaining commit history for pull requests. We can squash using GitHubs UI and the commit history can help reviewers. :)

lread:
Thanks for reviewing @martinklepsch!

I agree that adding SASS is a reasonable concern.  I started without it and found that I was spending too much time handcrafting hard to maintain css for asciidoc content. Using the very same  base that the asciidoctor team uses for styling html made for a much more pleasant and straightforward styling experience.

I expect that changing styling for asciidoc won’t happen very frequently and doesn’t absolutely need to be part of main build.

That said, if anyone else has other ideas, I’ve got an open mind.

Thanks for the CDN for font awesome, I’ll switch to that.  If the commit still looks heavy after that, I’ll see if I can split it to multiple commits to tell the story of the change better.

lread:
Hey @martinklepsch, I managed to switch over to CDN for font-awesome.  If you look at files changed, I think you'll find this much less overwhelming to review.

To give you an idea of the kind of css involved, look at the sass generated [asciidoc.css](https://github.com/cljdoc/cljdoc/blob/428f171cde3bce1215c25e15826a58752f1f7ab0/resources/public/cljdoc-asciidoc.css) and compare that to the relative simplicity of the sass source we would need to maintain [cljdoc-asciidoc.scss](https://github.com/cljdoc/cljdoc/blob/428f171cde3bce1215c25e15826a58752f1f7ab0/resources/sass/cljdoc-asciidoc.scss) and [settings/_cljdoc-asciidoc.scss](https://github.com/cljdoc/cljdoc/blob/428f171cde3bce1215c25e15826a58752f1f7ab0/resources/sass/settings/_cljdoc-asciidoc.scss)

martinklepsch:
I'll try to get around to test this tomorrow. 

Some notes to myself:

- explore if we could get rid of the ruby dependency by using https://github.com/sass/node-sass
- we have npm run stuff for the JS building, maybe we can combine these things using https://www.npmjs.com/package/concurrently
- maybe we can pull in tachyons etc via npm
- normalize.css seems to be duplicated (part of asciidoc but also part of tachyons)
- check that icons really improve things and are visually consistent with cljdoc's "style"
- consider combining all css 
- consider introducing a Makefile for the various make steps outlined in the "Running cljdoc locally" docs

lread:
Another option is to separate out the generation of the the asciidoc.css to a separate project. It might make seem more like a specialized task and make it less tempting to try to make it conform to tooling in cljdoc.

One thing I found helpful in my testing was cljoc-exerciser. It includes the entire asciidoctor manual which is a very thorough visual testbed.

lread:
One other option I thought of is to write a [custom asciidoctor backend](https://asciidoctor.org/docs/user-manual/#custom-backends).

Might be worth a deeper look to see how complicated it would be.  This might allow cljdoc to specify its asciidoctor formatting directly to elements via tachyons styles.

lread:
@martinklepsch, I think the important question is: What do we want asciidoctor documents to look like when presented by cljdoc?  Internal implementation details are much less important.

My strong preference is to support the visual richness of asciidoctor.  That's one of the main reasons why people write their docs in asciidoc format is it not?

lread:
I updated https://github.com/cljdoc/cljdoc-exerciser to allow use to more easily compare [CommonMark formatting](https://cljdoc.org/d/lread/cljdoc-exerciser/1.0.25/doc/document-tests/commonmark-features) with [AsciiDoc formatting](https://cljdoc.org/d/lread/cljdoc-exerciser/1.0.25/doc/document-tests/asciidoctor-features).

lread:
This is stale. I'll start fresh if when I take this up again.

PR #339, opened 2019-07-30
report on outdated dependencies

### LINKED ISSUE #333: Encourage ourselves to not let deps get too stale
**Current behavior**
When many cljdoc dependencies become old, we end up with the occasional big, riskier than we'd like, update.

**Desired behavior**
Trickle in dependency updates more frequently over time. 

**Action**
A [good little chat came](https://github.com/cljdoc/cljdoc/pull/331#issuecomment-514932934) up with the following idea:

> > The cljdoc circleci build could report, where possible, on outdated deps.  No failures, just a report via a simple step. Do you think that would encourage us to stay more up to date?
>
> If it's only printed during CI and doesn't show up on GitHub I'm not sure how many people will see it but if it's a step that is tracked as a "check" on GitHub I can see that being **very useful**.
>
> Screenshot 2019-07-25 at 09 35 16

lread:
⚠️ **might not merge until after #334 is merged to also deal with its deps and fully integrate  hardcoded version work**

Reports outdated dependencies for:
- clojure libraries
- node packages
- hardcoded JavaScript libraries

Setup as a job so that it will be clearly seen as part of GitHub
checks.

closes #333

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,6 +63,27 @@ jobs:
       - run: npm ci
       - run: npm run lint
 
+  # setup as job so that report is prominent under GitHub checks
+  dependencies-report:
+    docker:
+      - image: circleci/clojure:tools-deps-1.9.0.394-node
+        command: "/bin/bash"
+
+    working_directory: ~/repo
+
+    steps:
+      - checkout
+
+      - restore_cache:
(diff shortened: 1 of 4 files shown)


lread:
This is pretty darn stale now.  If I take this up again, I'll start fresh.

PR #340, opened 2019-08-02
Url fixups

### LINKED ISSUE #325: Images not rendered properly when testing locally
**Current behavior**
When testing [locally with cljdoc (non docker method)](https://github.com/cljdoc/cljdoc/blob/master/doc/running-cljdoc-locally.md#running-cljdoc-locally), I am finding that images are not showing up.

Note that `test-image-1.png` exists under `doc/`.

For the following markdown
```markdown
Images
--------
When bringing in a README.md for with `cljdoc.edn`, its will be referenced by `doc/README.md`. Are image links automatically fixed up?

Let's link to an image in a subdir with markdown syntax:
![alt for test image 1](doc/test-image-1.png "test image 1 text")

And again with html syntax:
<img src="doc/test-image-1.png" alt="alt for html test image 1"/>

```
Testing locally, I get:
image

For the following asciidoc
```asciidoc
This local image should work on github and cljdoc.  According to asciidoctor, images are resolved relative to the document.

Current value of imagesdir:

[subs=attributes+]
----
{imagesdir}
----

image:test-image-1.png[A local test image should appear here]

How about captions, do we display those nicely?

.Test image caption
image:test-image-1.png[A local test image should appear here]
```
Testing locally, I see:
image

**Desired behavior**
When deployed to cljdoc prod, I see what I expect to see locally.

Markdown:
image

asciidoc:
image

For full context see: https://github.com/lread/cljdoc-exerciser/tree/v1.0.3

lread:
**1) raw url now more generic**

The url used to reference images as raw content is no longer tied to
`raw.githubuserocntent.com` and now uses the more generic
`https://<git provider>/<owner>/<repo>/raw` syntax. This syntax works with
github, gitlab and bitbucket.

The sanitize query parameter required for raw svgs from github has been
preserved for github.

**2) fully qualified cljdoc urls now converted to root relative**

This supports local testing - we don't want to reach out to cljdoc.org when
running a local server for testing purposes.

**3) can now view images when testing locally**

After ingesting a local project's repo directory via the `--git` option, image
urls were definitively broken.

When cljdoc sees that the repo is a local directory, it will now try to find
your referenced images via a static htpp server running at
`http://localhost:9090`.  It will make simple raw fetches to this server without
github api adornments.  For example github fetch to:

```
https://github.com/cljdoc/cljdoc/raw/81f49c5f679d30dcdcad148de27b7b80501d315c/resources/public/cljdoc-logo-beta-square.png
```

becomes the following when running locally:
```
http://localhost:9090/resources/public/cljdoc-logo-beta-square.png
```

Startup your static web server on port 9090 in your project root using your
favorite technique. For example:
```
python -m SimpleHTTPServer 9090
```

For this first version port is not configurable and we don't entertain
supporting a local git server such as [adobe/git-server](https://github.com/adobe/git-server).

Closes #325


martinklepsch:
Hey Lee, I think this makes sense but misses some documentation, could we add some of this information to the "Running cljdoc locally" guide and maybe link to it when the localhost URI is used as base? 

I.e. When the repository URI is a filesystem URI we can log a warning that points to the respective section in the guide.

Let me know if this makes sense and thanks for working on this and adding tests too :)

lread:
Right you are @martinklepsch , I'll update the docs.

Other afterthoughts:
* I don't think I tested any of this when running locally with docker.  Will do - should all be good except for viewing images locally.
* I don't think this includes any url fixups in docstrings - will check on that.

src/cljdoc/util/fixref.clj
```diff
@@ -57,25 +56,42 @@
...
+
+(defn- raw-base-url [scm]
+  (or (local-raw-server scm)
+      (str (:url scm)
+           "/raw/"
+           (scm-rev scm) "/")))
+
+(defn- fix-image
+  [file-path src scm ]
+  (let [raw-base-url (raw-base-url scm)
```
martinklepsch, line 77: If we only need the `scm` arg to determine `raw-base-url` we could also just pass `(raw-base-url scm)` as an argument to this function. This might make it a little easier to read as well.

src/cljdoc/util/fixref.clj
```diff
@@ -57,25 +56,42 @@
...
+  (when (scm/fs-uri (:url scm)) "http://localhost:9090/"))
+
+(defn- raw-base-url [scm]
+  (or (local-raw-server scm)
+      (str (:url scm)
+           "/raw/"
+           (scm-rev scm) "/")))
+
+(defn- fix-image
+  [file-path src scm ]
```
martinklepsch, line 76:
```suggestion
  [file-path src scm]
```

src/cljdoc/util/fixref.clj
```diff
@@ -57,25 +56,42 @@
...
+  "This is overly inflexible for the first version.
+  When testing locally we point to simple static http server."
+  [scm]
+  (when (scm/fs-uri (:url scm)) "http://localhost:9090/"))
+
+(defn- raw-base-url [scm]
+  (or (local-raw-server scm)
+      (str (:url scm)
+           "/raw/"
+           (scm-rev scm) "/")))
```
martinklepsch, line 73: Reviewing this code I was a little concerned that `raw-base-url` is different than what has previously been done here
CleanShot 2020-06-11 at 22 40 30@2x

It seems that instead of using `https://raw.githubusercontent.com/` we now use the `/raw/` URL segment to access raw versions of the files. If these can be used interchangeably then that change is fine with me, just trying to understand if/why something changed here. 🙂

lread: Thanks for the concern, this is certainly a rusty PR and merits review!  This change is described in the PR message under "1) raw url now more generic".

src/cljdoc/util/fixref.clj
lread: Yeah, good point!  Will address.

src/cljdoc/util/fixref.clj
martinklepsch: Duh, should have referred back to the PR description before asking haha 😄 Sounds good then!

lread:
Ooops, accidental push... closing this one will reopen new PR shortly.

PR #331, opened 2019-07-24
Upgrade cljdoc dependencies

lread:
resolves #328

PR #346, opened 2019-09-06
Upgrade pedestal

martinklepsch:
Follow up to #331 #345 

When wanting to bind to anything other than localhost pedestal now requires the `::host` key to be set explicitly. We previously didn't do that, and it's probably a little miracle that everything worked anyways. 

The issue with setting `::host` is that we want to set it to `localhost` when running things in the REPL but inside Docker it needs to be set to `0.0.0.0`. This may require us to introduce another profile among `prod` and `default`. 

Other suggestions/ideas very welcome.

https://github.com/cljdoc/cljdoc/commit/7cace9c8fa46778e7b3b9a27bce5140d9a1676ef makes things work in Docker but also means that `0.0.0.0` is used as host for local development. Not ideal I guess.


lread:
Hey @martinklepsch, congrats on figuring out the deploy problem! 👍

lread:
Looking into the mystery of how things worked before, it seems that the [default host for jetty is 0.0.0.0](https://www.eclipse.org/jetty/documentation/current/quickstart-config-what.html#intro-jetty-configuration-connectors):

> You can configure a host either as a host name or IP address to identify a specific network interface on which to listen. If not set, or set to the value of 0.0.0.0, the connector listens on all local interfaces.

lread:
Superseded by #494

Issue #336, opened 2019-07-28
API diffs between different versions?


danielcompton:
It would be neat if we could diff two versions of a library to show the API changes between them. The first use-case I have is just checking what has changed in a library between the version I'm on and the version that I'm upgrading to. 

The way I imagine it working would be:

* If a function/var has been added, it would be marked with a green +
* If a function/var has been removed, it would be marked with a red -
* If a function/var has had changed, it would be marked with a blue ~

The + and - cases are relatively easy, but it's not super clear what the semantics of ~ should be. What does it mean for a function to change, and how would we know? Would you want to mark ~ whenever the function source code or docstring changed, even if it returned the same result? Would you want to distinguish between breaking and growing changes? How would you tell?

As you can tell, this gets to be a little bit speculative when you start to dig into what it would really mean. Perhaps a useful first pass would be to support the + and - cases and make a note that changes to functions aren't covered?

lread:
Yes, I think this is a great idea, too!  

When you say "a little bit **spec**ulative" did you intend anything there? 😉 

For ~, I think we could start with detecting function signature changes.

danielcompton:
Heh, yeah spec was definitely on my mind there. The challenge with only detecting function signature changes is that it may have false negatives where the function's semantics have changed but the arguments have stayed the same. Diffing docstring and function signature changes would probably get you 90% of the way there, and it would still be a big improvement over the status quo.

lread:
My personal interest is in breaking changes, but knowing about new functions will also be useful.

API diffing will be very helpful but never entirely conclusive.  An unchanged function can call a changed function.  A changed function might have simply been refactored. 

Looking at a function's tests might be helpful. If a function's tests have changed, that might indicate behavior has changed, or might not.  

For ambiguous cases, somehow knowing the API author's intent would be of great value.

I think starting small and working on this incrementally makes very good sense.

martinklepsch:
We can infer some breakage from arglists. Eg. if there’s been 3 args but now the function takes 4 and has no 3-arg arity anymore.

Changes that aren’t breaking could be just marked as “change” and I think that would still be useful as it happens within the conceptual space of Clojure (functions, arguments, docstrings) rather than text files.

Excited to see this being discussed more :)

martinklepsch:
FYI, I started working on this, will open a PR later today. Some things that we can do beyond just diffing function arguments and docstrings:

- Because we have Git revisions we can link to the diff page on the respective Git service. 
[...]

PR #347, opened 2019-09-10
refactor data loading to enable loading data for multiple artifacts

martinklepsch:
prep work for #336

### DIFF
--- a/src/cljdoc/server/pedestal.clj
+++ b/src/cljdoc/server/pedestal.clj
@@ -41,6 +41,13 @@
             [io.pedestal.interceptor :as interceptor]
             [io.pedestal.http.ring-middlewares :as ring-middlewares]))
 
+(defn get-artifact-data
+  "Return the artifact data from the contexts `::artifacts` key for the provided version-entity `v-ent`."
+  ([ctx v-ent]
+   (get-in ctx [::artifacts (select-keys v-ent [:group-id :artifact-id :version])]))
+  ([ctx v-ent k]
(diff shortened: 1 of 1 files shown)


src/cljdoc/server/pedestal.clj
```diff
@@ -104,43 +113,53 @@
...
                 :cljdoc/index   (index-pages/full-index (::releases ctx)))))])
 
+(defn seed-artifacts-keys
+  "An interceptor to prepare `::artifacts` key for [[artifact-data-loader]]."
+  [data-requirements]
+  (interceptor/interceptor
+   {:name ::seed-artifact-keys
+    :enter (fn seed-artifact-keys-inner [ctx]
+             (let [params (-> ctx :request :path-params)]
+               (assoc-in ctx [::artifacts (select-keys params [:group-id :artifact-id :version])] data-requirements)))}))
```
martinklepsch: The idea here is that this function would later on be extended to process `ctx` in more interesting ways, potentially resulting in different `::artifacts` values.

This would most likely be URL or query params that are provided through the `:request`. I wanted to keep this separate from the interceptor that's actually loading the data because this way the loading interceptor does not need to know about route names, where path or query params are stored and all that. Hope that makes sense.

lread:
So if I understand correctly (which I might not, please correct me where I don't!):
* you've localized all data loading to one place abstracting it away from interceptor code
* abstracted fetching data from the context 
* defined a scheme for a route-resolver to state what data it needs

These seem like good things.

It might be worth playing out the api diff feature to see how it works out.  What would change?

lread:
Chat with @martinklepsch; we think this work is stale enough to close for now.

Issue #308, opened 2019-06-04
Search results improvements


holyjak:
Improve results produced by the new search introduced by #85.

Report problems and bad/suboptimal results here. (Check the Known Problems below first, please!)

## Known Problems

* Include download count in the prioritization to show more popular libraries first - see #359 
* Searching for "clojure spec" does not show the library in the presented top results ("org.clojure/spec" does)
* Should prioritize "standard" versions over weird one, e.g. 0.6.6 over 0.6.7-AlexVPopov_patch_1-SNAPSHOT (?How often does this happen, is it worth the extra complexity?)

martinklepsch:
Not a big deal but one thing I noticed is that searching for `metosin json` doesn't return `metosin/jsonista`.

KingMob:
I'm not sure if search is set up for non-Clojars sources, but I couldn't find tools.reader using either `tools.reader` or `org.clojure/tools.reader`.

holyjak:
Thanks a lot, @KingMob ! It will be fixed in a few minutes. We fetched only the first 10 instead of all 64 artifacts.

gphilipp:
I'm searching for `jackdaw`, but the only result that pops up is some version built off a branch https://cljdoc.org/d/fundingcircle/jackdaw/0.6.7-AlexVPopov_patch_1-SNAPSHOT, instead of the latest release: https://cljdoc.org/d/fundingcircle/jackdaw/0.6.6/doc/readme

holyjak:
I guess the problem is that 0.6.7* > 0.6.6 and the code doesn't really look at the versions to filter out "weird ones". I will look into it, eventually.

escherize:
Firstly, cljdoc is a really cool tool that I am fond of!

I think it should be a goal that when searching for e.g. `lacinia`, the "main library" (aka `com.walmartlabs.lacinia 0.34.0`) shows up first, but what I am seeing is:

- toolbelt-lacinia 0.0.1
- lacinia-gen 0.1.1-SNAPSHOT
- lacinia2datomic 0.1.0-SNAPSHOT
- ... [snip] ...
- `com.walmartlabs.lacinia 0.34.0`

Do you agree that this should be considered a bug?

holyjak:
[...]

PR #351, opened 2019-09-16
Add com.turtlequeue to group IDs searched on Maven Central

nha:
Following discussion on this issue: https://github.com/cljdoc/cljdoc/issues/308#issuecomment-530584697

Notes:
 - I did not include the quotes in the search as they do not seem useful
```
http://search.maven.org/solrsearch/select?q=g:%22org.clojure%22+OR+g:%22com.turtlequeue%22&rows=200
```
returns the same as
```
http://search.maven.org/solrsearch/select?q=g:org.clojure+OR+g:com.turtlequeue&rows=200
```
- the `description` will be wrong https://github.com/cljdoc/cljdoc/blob/1eaaac904b275ebe581f1775a3c4b5bb44e43bdd/src/cljdoc/server/search/artifact_indexer.clj#L36

### DIFF
--- a/src/cljdoc/server/search/artifact_indexer.clj
+++ b/src/cljdoc/server/search/artifact_indexer.clj
@@ -6,7 +6,8 @@
     [clojure.java.io :as io]
     [clj-http.lite.client :as http]
     [clojure.tools.logging :as log]
-    [cheshire.core :as json])
+    [cheshire.core :as json]
+    [clojure.string :as string])
   (:import (org.apache.lucene.analysis.standard StandardAnalyzer)
            (org.apache.lucene.index IndexWriterConfig IndexWriterConfig$OpenMode IndexWriter Term IndexOptions)
            (org.apache.lucene.document Document StringField Field$Store TextField FieldType Field)
--- a/src/cljdoc/server/search/artifact_indexer.clj
+++ b/src/cljdoc/server/search/artifact_indexer.clj
@@ -36,10 +36,17 @@
                :description (str "Clojure Contrib library " g "/" a)
                :origin :maven-central}))))
 
+(def ^:private maven-artefacts ["org.clojure"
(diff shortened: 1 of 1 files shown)


martinklepsch:
@nha this looks reasonable but the build is failing due to a missing `clojure.string` require.

To make review easier it's also nice if you could say how you tested this and how others can validate the changes you made. :)

martinklepsch:
@holyjak I'll leave it to you to merge/accept this PR since you're most familiar with this part of the codebase.

nha:
re: testing - I only I tested in the REPL that the result of `load-maven-central-artifacts` was coherent with the previous result

src/cljdoc/server/search/artifact_indexer.clj
```diff
@@ -36,10 +37,17 @@
                :description (str "Clojure Contrib library " g "/" a)
                :origin :maven-central}))))
 
+(def ^:private maven-artefacts ["org.clojure"
```
holyjak: "artefacts" is wrong, correct would be `maven-groups`

holyjak: "artefacts" is misleading, correct would be maven-groups

holyjak:
Perhaps also change https://github.com/cljdoc/cljdoc/blob/1eaaac904b275ebe581f1775a3c4b5bb44e43bdd/src/cljdoc/server/search/artifact_indexer.clj#L36 to something like 

```
:description (str "Maven Central Clojure library " g "/" a)
```

?

src/cljdoc/server/search/artifact_indexer.clj
nha: fixed in 
81ce58c

holyjak:
thank you!

PR #353, opened 2019-09-23
Switch to new analyzer

martinklepsch:
@lread has done a lot of great work in his https://github.com/lread/cljdoc-analyzer project, forking codox and combining it with the various tools we have built around it. The basic idea behind this change is that the analyzer itself may also be interesting to other people in the Clojure community that want to programmatically extract information from Clojure code.

This PR makes the necessary modifications to switch over to the new analyzer, with some details to be figured out. So far most things seem to work as expected.


martinklepsch:
So I did a bit more testing of this. The steps I've taken are:

1. Build a docker image `cd ops/docker; make image; cd -`
2. Run the docker image `docker run --rm -p 8000:8000 -v data:/app/data cljdoc/cljdoc:0.0.1511-new-cljdoc-analyzer-29a4de2`
3. Go to http://localhost:8000/d/lilactown/hx/0.5.3 & hit "build"
4. Go to http://localhost:8000/d/funcool/promesa/5.1.0 & hit "build" 

Both builds seem to currently fail. Steps we could take to understand where it fails:

- Run outside docker & repeat steps above (`clj -A:cli run`)
- Run https://github.com/cljdoc/cljdoc-analyzer on these projects individually, there is a test for `hx` https://github.com/cljdoc/cljdoc-analyzer/commit/d7cb539620dc2e2b910679f16931b7466dc320d5 so I'm thinking that that's probably not the culprit here.

PR #354, opened 2019-10-01
Add a cljdoc developers guide

### LINKED ISSUE #344: Add a “for developers” guide
We have a good overview `CONTRIBUTING.adoc` and some specific details in `doc/running-cljdoc-locally.md` but a general place to guide developers in the specifics is missing.  Recent and upcoming changes merit a `doc/for-developers.adoc`. TBD if this doc replaces or references `doc/running-cljdoc-locally.md`.

lread:
A spot for us to add technical developer specific notes.

Of note:

1. I added my best guesses at some things/philosophies, please review with a
critical eye.

2. I added a system overview diagram - something I would have liked when coming
aboard. I think I got it right. Let me know.

3. I experimented with ascii art for a system overview diagram. (As an aside, it
looks quite nice when asciidoctor's diagramming feature is turned on.)
Let know what you think. The idea was to use a diagram tech that is accessible
to everyone and also tracks under version control.

3. I reworked the running locally doc to talk to its two main audiences more
clearly.

Closes #344


lread:
Hey just fyi, here's what the ascii art looks like when rendered by [asciidoctor diagram](https://asciidoctor.org/docs/asciidoctor-diagram/) using svgbob:

image

Not suggesting we enable this now, just found it of interest.

doc/cljdoc-developer-technical-guide.adoc
```diff
@@ -0,0 +1,94 @@
...
+  --|--   publish    │  │   project    │  │            for api   │  │ cljdoc api  │  │
+    |   ─────────────┼─>│   artifact   ├──┼──────────────────────┼─>│  analysis   │  │
+   / \    library    │  │              │  │           analysis   │  │     job     │  │
+  library            │  └──────────────┘  │                      │  │             │  │
+  author             └───────────────┬────┘                      │  └──────────┬──┘  │
+                                     │                           │     ^       │     │
+                                     │     discovers             └─────┼───────┼─────┘
+                                     └───────────────────┐       launch│       │api
+                                         new library     │          api│       │analysis
+                                                         │     analisys│       │result
```
martinklepsch:
```suggestion
                                                         │     analysis│       │result
```

martinklepsch:
Oh and maybe add that new doc file to the index in `doc/cljdoc.edn` 😅

doc/cljdoc-developer-technical-guide.adoc
lread: tx, will correct!

lread:
Hi @martinklepsch! Thanks for finding some time to review!

I gotta say, ascii art is not the easiest to edit. I fiddled with various emacs modes but did not find something I understood.  All I really wanted was an overwrite/canvas-like mode that did not shift the text, but hey, my emacs skills are still budding.

I tried online tools [textik](https://textik.com) and [asciiflow](http://asciiflow.com/) found them good but not great. I tried [MonoDraw](https://monodraw.helftone.com/), a dedicated ascii art drawing app on macOS which was pretty good and I used it for my final drawing.  In the end I tweaked the drawing by hand slightly to follow svgbob conventions because I was curious to see what svgbob would render.  

Note that because MonoDraw is macOS only and not free ($10USD) it won't be a solution for everyone. 

SVG might be a good compromise! Less accessible but more expressive and likely easier to edit. I've used the open source InkScape with some success in the past for SVG drawings. Note that InkScape currently has some frustrating bugs on macOS that sometimes render windows offscreen. This has me moving over to linux when I use InkScape but I think they are working on this issue for their 1.0 release.

If we do use SVG, we should probably mention a tools contributors can use. Do you have any favorites?  And while I'm thinking out loud, accessibility and editability are probably more important than traceability under version control. So even a png format would be fine as long as the original editable format was committed to version control as well.  (And that format was accessible to contributors).

I know that the Cognitec team uses OmniGraffle for their drawings. I've also used this in the past and very much like it.  But OmniGraffle starts at $149USD and is therefore not accessible.

As to the detail of the drawing, I knowingly left out some finer points to keep brush strokes broad, and the diagram easy to digest. But I do wonder if I might have been a bit more detailed with a finer grained drawing tech like SVG.

martinklepsch:
I have some friends who used [draw.io](https://www.draw.io) heavily and like it. Since it's web based and free it's probably one of the most accessible options out there.

lread:
Hmmm!  [draw.io](https://www.draw.io) looks very nice! 😃 I see they also have [desktop apps](https://about.draw.io/integrations/) for those that prefer that.

So seems like the following would make sense? Folks would commit their original .drawio file and a web friendly format (.png?) to reference from their doc.

lread:
Wow, draw.io id great, thanks for sharing! I switched over.

PR #357, opened 2019-10-14
Support also JSON, EDN for /versions/*

### LINKED ISSUE #350: Provide API to get all available versions for a project
[Dash](https://www.kapeli.com/dash) has been using Clojars' API to figure out what versions are available (In Dash you can install documentation for older versions). 

E.g. https://clojars.org/api/artifacts/cljdoc/cljdoc

The issue is that we now also host documentation for various artifacts coming from Maven Central (core.async, Clojure itself, etc.) and for these artifacts this API doesn't work.

**We will need to provide our own endpoint to return the versions for those artifacts. The data returned can be based on our internal database.**

Perhaps a good approach would be to extend the `index-routes` below to handle additional content types (JSON):
```clj
(defn index-routes []
  #{["/versions" :get nop :route-name :cljdoc/index]
    ["/versions/:group-id" :get nop :route-name :group/index]
```

cc @holyjak feel free to add to this if you have any further thoughts.

holyjak:
so that Dash can use it to fetch the available versions of a document.

(Impl.: Transform versions seq into the versions-tree of
group->artifacts->versions, all sorted, use it everywhere.)

Fixes #350

### DIFF
--- a/src/cljdoc/render/index_pages.clj
+++ b/src/cljdoc/render/index_pages.clj
@@ -9,21 +9,15 @@
             [clojure.spec.alpha :as spec]
             [version-clj.core :as v]))
 
-(spec/fdef sort-by-version
-  :args (spec/cat :version-entities (spec/coll-of :cljdoc.spec/version-entity)))
-
-(defn sort-by-version [version-entities]
-  (sort-by :version #(- (v/version-compare %1 %2)) version-entities))
-
 (spec/fdef artifact-index
   :args (spec/cat :artifact-entity :cljdoc.spec/artifact-entity
                   :versions (spec/coll-of :cljdoc.spec/version-entity)))
 
 (defn artifact-index
-  [artifact-entity versions]
-  (let [matching? #(= (:artifact-id artifact-entity) (:artifact-id %))
-        matching  (filter matching? versions)
-        others    (group-by :artifact-id (remove matching? versions))
+  [{:keys [group-id artifact-id] :as artifact-entity} versions-tree]
+  (let [matching  (get-in versions-tree [group-id artifact-id])
+        others    (-> (get-in versions-tree [group-id])
+                      (dissoc artifact-id))
         btn-link :a.dib.bg-blue.white.ph3.pv2.br1.no-underline.f5.fw5
         big-btn-link :a.db.link.blue.ph3.pv2.bg-lightest-blue.hover-dark-blue.br2]
     (->> [:div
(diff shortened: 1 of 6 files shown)


src/cljdoc/server/pedestal_util.clj
```diff
@@ -61,3 +61,15 @@
    {:name ::html
     :enter (fn html-render-inner [context]
              (ok-html context (render-fn context)))}))
+
+(defn body
```
holyjak: Can we think of a better name than `body`? Perhaps `ok-html-or-data`?

src/cljdoc/render/index_pages.clj
```diff
@@ -99,17 +98,27 @@
...
+            [:h2 group-id [:span.gray.fw3.ml3.f5 "Group ID"]]
             [:div.nl2.nr2
-             (for [[_a-id versions-for-artifact] (group-by :artifact-id versions-for-group)
-                   :let [latest (first (sort-by-version versions-for-artifact))]]
+             (for [[_a-id versions-for-artifact] groups-artifact-id->versions
+                   :let [latest (first versions-for-artifact)]]
                (artifact-grid-cell latest))]])]]
        (layout/page {:title "all documented artifacts — cljdoc"})))
+
+(defn versions-tree
```
holyjak: It doesn't seem optimal to have the `versions-tree` here in index-pages but is there a better place for it somewhere?

src/cljdoc/server/pedestal.clj
```diff
@@ -98,11 +100,12 @@
 
                 :cljdoc/index
                 (assoc ctx ::releases (storage/all-distinct-docs store))))})
-   (pu/html (fn [ctx]
+   (pu/body (fn data [ctx] (-> ctx ::releases index-pages/versions-tree))
+            (fn html [ctx versions-tree]
```
martinklepsch: I think the way this is done usually is that you just set `:response :body` to whatever data would be returned as JSON or EDN and then you provide an extra function that turns that into an html string. 

There's some more reference material here: http://pedestal.io/guides/hello-world-content-types#_returning_to_content_types

martinklepsch: Maybe if implemented properly we could refactor a bunch of our helpers that way.

src/cljdoc/server/pedestal_util.clj
martinklepsch: This could be `coerce-body` as per the reference link I posted in the other comments.

src/cljdoc/render/index_pages.clj
martinklepsch: Not too concerned about this. Maybe it could be moved into `cljdoc.util` but I don't really mind it being here. One thing that might be nice though: move the `sort-by-version` helper closer to this function since it's the only one using it.

holyjak:
Hi Martin, sorry that it took so long time! These two weeks have been crazy at work.

Thank you for the suggestions! I have changed it so that the top+level interceptor `pu/coerce-body-conf` (formerly pu/coerce-body) now takes an optional `transformer` function taking `[request content-type body]` that can transform the body prior to it being stringified. The bottom-level interceptor then simply adds extracts the data and adds them `:response :body` as suggested. It is certainly nicer now (though I have to pass the transfomrer through three levels :-( ).

Any more comments? (I still have to test it...)

holyjak:
I have now tested it and it seems fine.

PR #359, opened 2019-10-29
Search score by download counts

holyjak:
TODO Adjust the query structure / weights so that the effect is neither too weak nor does it push forward worse matches. Tricky! See `boost-by-downloads` at https://github.com/cljdoc/cljdoc/pull/359/files#diff-2ce74948e9ce83d10fbff231a0737ef7R121

via #308 

Note: This is how clojars incorporate download counts

### DIFF
--- a/resources/public/opensearch.xml
+++ b/resources/public/opensearch.xml
@@ -7,7 +7,7 @@
     <Url type="text/html" template="https://cljdoc.org/search">
         <Param name="q" value="{searchTerms}"/>
     </Url>
-    <Url type="application/x-suggestions+json" template="https://cljdoc.org/suggest">
+    <Url type="application/x-suggestions+json" template="https://cljdoc.org/api/search-suggest">
         <Param name="q" value="{searchTerms}"/>
     </Url>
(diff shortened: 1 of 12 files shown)


lread:
Hiya @holyjak! I am doing a maintenance pass on cljdoc, and the last set of deps I have not yet updated yet is lucene (current version is 9.0.0). 

You ok with me going ahead with the update? (Our API usage of lucene seems intact, I might add a few tests to check for any differences in results).

lread:
Hmmm... I just noticed that clojars is also using Lucene. 
We can probably get inspiration from their hard work over there!

holyjak:
Hi! Sorry for the late reply. Absolutely go ahead with the update!

And I am aware of the work at cljdoc and wish to steal from it. I just have so much on my plate that I an unable to prioritize it (also because I am afraid it would need a few days of exclusive focus to get done and I typically just have few hours here and there 😭) See https://github.com/cljdoc/cljdoc/issues/308#issuecomment-1021599750

lread:
Thanks @holyjak I learned a lot from this PR!
It is now superseded by my little effort in #588, so closing this one.

holyjak:
❤️

PR #361, opened 2019-11-01
Refactor how HTML body is created in coerce-body helpers

martinklepsch:
Looking at that code to understand the search issue I had a bit of a hard time understanding what the `transformer` is actually doing. It is passed down very far and in many places augmented/changed slightly making overall program flow a little hard to understand (to me). 

I'm pretty sure there are still issues with this PR, particularly the `show-build` interceptor may need some touching up, but I wanted to open this just to get your thoughts on the general refactoring @holyjak.

### DIFF
--- a/src/cljdoc/server/pedestal.clj
+++ b/src/cljdoc/server/pedestal.clj
@@ -259,9 +259,7 @@
     :enter (fn build-show-render [ctx]
              (if-let [build-info (->> ctx :request :path-params :id
                                       (build-log/get-build build-tracker))]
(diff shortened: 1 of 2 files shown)


holyjak:
Yes, it is simpler and it seems unlikely that we would need any other "transformer" than for HTML. We should perhaps rename `coerce-body-conf` to something more matching? Or perhaps just `coerce-body`?

martinklepsch:
@holyjak please merge if you think this is good. I skipped the renaming for now for the lack of any good name ideas 😛

PR #362, opened 2019-11-01
Add asciidoctor example to badges

SevereOverfl0w:


martinklepsch:
Thanks Dominic!

I added you to the cljdoc org now so you can also merge pull requests and push new commits. 🚀

PR #364, opened 2019-11-05
Fix badge routes

### LINKED ISSUE #348: Ensure that /badge endpoint will 404 with more than one slash in project ID
My library is there: https://cljdoc.org/d/com.turtlequeue/clj-client/0.0.7
It seems I should be able to get a badge https://cljdoc.org/badge/com.turtlequeue/clj-client/0.0.7

NOTE: there is no SCM configured - I am assuming this may be the cause here

Also I noticed it does not appear in the search but it may be unrelated ( https://github.com/cljdoc/cljdoc/issues/308#issuecomment-530584697 )

lins05:
This shall fix #348 .

In https://github.com/cljdoc/cljdoc/issues/348#issuecomment-531135828 it's suggested to fix it by use two separate routes:

```clojure
["/badge/project-id" :get nop :route-name :badge-for-project]
["/badge/group-id/artifact-id" :get nop :route-name :badge-for-project]
```

However I think using constraints are slightly better because pedestal requires each route to have a unique name, which means we need to duplicate the `:badge-for-project` entry in the `route-resolver` function:

https://github.com/cljdoc/cljdoc/blob/f08a21696d887c0c357a3f75338167c1d5b18d6b/src/cljdoc/server/pedestal.clj#L431-L433

(It's a pity that pedestal doesn't support capturing path params with regular expressions, btw)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ node_modules
 .envrc
 resources-compiled
 pom.xml
+test-data/
--- a/src/cljdoc/server/routes.clj
+++ b/src/cljdoc/server/routes.clj
@@ -52,17 +52,14 @@
     ["/shortcuts" :get nop :route-name :shortcuts]
     ["/sitemap.xml" :get nop :route-name :sitemap]})
 
-(def project-path
-  "Ensure at most one slash.
-  See https://github.com/cljdoc/cljdoc/issues/348"
-  #"^[^/]+(/[^/]+)?$")
-
 (defn utility-routes []
   #{["/jump/release/*project" :get nop :route-name :jump-to-project]
     ["/badge/*project"
      :get nop
      :route-name :badge-for-project
-     :constraints {:project project-path}]})
+     ;; Ensure at most one slash.
+     ;; See https://github.com/cljdoc/cljdoc/issues/348
+     :constraints {:project #"^[^/]+(/[^/]+)?$"}]})
 
 (defn routes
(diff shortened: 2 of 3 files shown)


lins05:
ping @martinklepsch

martinklepsch:
Sorry for the delay on this, unfortunately my computer has some display issues so I’m not going to get to merge this before Tuesday. 

I think I’d suggest to just add the regex inline with a small comment similar to the `project-path` docstring, then we don’t need to explicitly name this and save a couple of lines :)

martinklepsch:
Oh and — completely forgot — congrats on your first contribution! Even with tests and all. If there was a scoring system this would score pretty high I’d say 😄

I’d say I send you some stickers but I’m afraid I have run out of them. (By which I really mean I misplaced the fee remaining ones that I still had.) I still have loads of other nice stickers though so email me your address if you’re into that :)

martinklepsch:
Hey @lins05, just wanted to say thanks again for your contribution, I just realized yesterday that you probably found cljdoc via our conversation on Reddit. 

If you have any feedback regarding developer-focused docs for cljdoc please let us know! And if you would like to contribute further I'd be happy to hear more about why you contributed in the first place and maybe see if there's something in cljdoc that matches this motivation :)

lins05:
@martinklepsch I contributed mainly to get my hands dirty on a real world clj project :) 

>  if there's something in cljdoc that matches this motivation :)

Well, anything that could help the project would be ok :D

PR #368, opened 2019-11-25
Fix #350 /versions API include unbuilt

### LINKED ISSUE #350: Provide API to get all available versions for a project

holyjak:
Build on #357 to return all versions.

Also improves indexing to get all versions of Maven Central artifacts and to fetch descriptions for those that have one.

### DIFF
--- a/src/cljdoc/server/search/artifact_indexer.clj
+++ b/src/cljdoc/server/search/artifact_indexer.clj
@@ -3,11 +3,11 @@
     [cljdoc.spec :as cljdoc-spec]
     [clojure.edn :as edn]
     [clojure.spec.alpha :as s]
+    [clojure.string :as str]
     [clojure.java.io :as io]
     [clj-http.lite.client :as http]
     [clojure.tools.logging :as log]
-    [cheshire.core :as json]
-    [clojure.string :as string])
+    [cheshire.core :as json])
   (:import (org.apache.lucene.analysis.standard StandardAnalyzer)
            (org.apache.lucene.index IndexWriterConfig IndexWriterConfig$OpenMode IndexWriter Term IndexOptions)
            (org.apache.lucene.document Document StringField Field$Store TextField FieldType Field)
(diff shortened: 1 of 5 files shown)


holyjak:
@martinklepsch I'd appreciate your feedback. Especially

1. How much should I use Spec? 
2. Where to put the functions? Perhaps we should create a new `data.documents` namespace and put all of `available-docs-denormalized, versions-data, index-pages/versions-tree` there?

src/cljdoc/server/search/artifact_indexer.clj
```diff
@@ -22,41 +22,117 @@
...
+    (->> (fetch-maven-docs (str "g:" group-id))
+         (update-grp-version-count! group-id)
+         (map maven-doc->artifact)
+         (mvn-merge-versions)
+         (pmap add-description!))))
+
+
+(defn load-maven-central-artifacts
+  "Load artifacts from Maven Central - if there are any new ones (or `force?`)
+  NOTE: Takes ± 2s as of 11/2019"
```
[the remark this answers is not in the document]
holyjak: Never. I just thought it might be useful to know for us devs whether it takes milis or minutes.

holyjak:
Hi Martin, thank you!!!

>     2. I guess somewhere under `search.*` would also work? No strong opinions here I think.

I don't think so, because versions have nothing to do with search (speaking about the current https://cljdoc.org/versions/ HTML & JSON output, without the `?all=true`). This has to do with data transformations and in particular with "documents" so `data.documents` seems a better match. Though there are certainly better names we can come with. `data.versions`?

>     3. That sounds fine but I didn't completely understand why this cache exists in the first place. If reading from the Lucene index takes too long and we need caching then sure, lets cache it.

It takes over 5s _on my PC_. I don't know, perhaps we don't care about that and it is good enough to do this and incur the delay on each request rather than adding the complexity of caching?

>  I think ideally the state of that cache would somehow be captured in our system. Maybe as an implementation of `ISearcher` that also caches its results. I think that could be pretty nice.

Could you please elaborate on the benefits and implementation of this?

src/cljdoc/server/search/search.clj
```diff
@@ -185,6 +193,39 @@
               :doc (.doc (:searcher %) (.-doc h))})
            (:score-docs %))))))
 
+(defonce all-docs-cache (atom nil)) ;; As of 11/2019 it takes ± 2.6MB (well, when stringified)
+
+(defn all-docs
+  "Returns all the documents in the Lucene index, with all the versions.
+  NOTE: Takes a few seconds."
+  [^String index-dir]
```
martinklepsch: Could we extend this to already take group and artifact IDs? Then we would not have to filter the entire data in `available-docs-denormalized`

holyjak: It could but then we should create 3 fns - for all, for group, for artifact. And maybe give up on caching and just read the raw index every time, b/c it would be too complicated for all 3 cases?

What is the gain you seek? Performance? Or simpler code by making Lucene do the filtering? But the complexity just moves to here it doesn't really disappear I think.

I do return all documents because I believe this is what we do for the normal `/versions/*` requests as well. No? However I am open to anything...

holyjak:
I propose to make an effort to get this out ASAP so Dash can start using it and then eventually add improvements later. We could either merge it as is or I could drop the caching code if the state worries you. Thoughts?

src/cljdoc/server/search/search.clj
martinklepsch: > What is the gain you seek? Performance? Or simpler code by making Lucene do the filtering? But the complexity just moves to here it doesn't really disappear I think.

I was wondering if maybe if we constrain at the Lucene level the more common queries (to get group/artifact versions) would be much faster than the current `all-docs` implementation.

martinklepsch:
I'm ok with that. In general you can totally merge stuff if you feel like it's ready 🙂  Maybe add a `TODO` comment touching on what we discussed & linking to the conversation here.

PR #370, opened 2019-12-12
Add a catch-all error interceptor

### LINKED ISSUE #367: Implement catch-all to avoid Pedestal "stylobate" catch-all
Pedestal's default catch-all interceptor logs the entire content which can get big. Huge log lines seem to impact responsiveness of the server so we should probably implement our own handling of these unhandled exceptions. For now a simple log line with the exception class and message might be enough.

hindol:
Closes #367.


lread:
Not sure if we need this one? Just merged #512.

lread:
Gonna assume #512 did the trick, we can reopen this one if we learn that I'm wrong.

Issue #373, opened 2020-01-08
Not loading namespace with special symbols in it


wilkerlucio:
Hello, I recently tried to publish the docs for this library: https://cljdoc.org/d/com.wsscode/edn-json/1.0.0/api/com.wsscode.edn%3C-%3Ejson

But the namespace is not loading for some reason, maybe because of the `<->` in the namespace name.

PR #385, opened 2020-04-15
Handle namespaces with special symbols in its name

kloimhardt:
Adresses issue #373 

- [X] Make a first quick hack for the special symbol "<->" of issue #373
- [X] Ask the community about the best way to proceed
- [x]  Solve for arbitrary special symbols in an idiomatic manner

### DIFF
--- a/src/cljdoc/server/pedestal.clj
+++ b/src/cljdoc/server/pedestal.clj
@@ -39,7 +39,8 @@
             [io.pedestal.http :as http]
             [io.pedestal.http.body-params :as body]
             [io.pedestal.interceptor :as interceptor]
-            [io.pedestal.http.ring-middlewares :as ring-middlewares]))
+            [io.pedestal.http.ring-middlewares :as ring-middlewares]
+            [lambdaisland.uri.normalize :as normalize]))
 
 (def render-interceptor
   "This interceptor will render the documentation page for the current route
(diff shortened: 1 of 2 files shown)


kloimhardt:
This PR is my first attempt to contribute to cljdoc (took some hours to come up with this one liner). Before carrying on, I'd like to ask:

1) As I do not have any experience with pedestal/interceptors, is this the right place to do the string replacement?
2) Which library would be best to do general special symbol replacements?

martinklepsch:
Hey @kloimhardt, thanks for contributing, this is an excellent first issue to contribute! 

1. I think the place you picked looks reasonable for now. Maybe link to this issue from there for later reference but other than that this looks fine. There is a general question if we should just process all `path-params` with the function below.
2. I think [`lambdaisland.uri.normalize/percent-decode`](https://cljdoc.org/d/lambdaisland/uri/1.2.1/api/lambdaisland.uri.normalize#percent-decode)might be a good option here. We already depend on that library so you should be able to just require it.

kloimhardt:
Ready for merging into master, fixes #373. For the topic of "processing of all path-params", I would rather look into this in th future, once this PR has been shown not to break something obvious.

src/cljdoc/server/pedestal.clj
```diff
@@ -52,7 +53,12 @@
   (interceptor/interceptor
    {:name  ::render
     :enter (fn render-doc [{:keys [cache-bundle] :as ctx}]
-             (let [path-params (-> ctx :request :path-params)
+             (let [pp (get-in ctx [:request :path-params])
+                   path-params
+                   (cond-> pp
+                     ;; fixes #373
```
martinklepsch, line 59:
```suggestion
                     ;; fixes https://github.com/cljdoc/cljdoc/issues/373
```

PR #386, opened 2020-04-16
add version 0.6.1 to muuntaja URL in running-cljsdoc-locally.md

kloimhardt:
This is necessary because the URL without 0.6.1 defaults (upon pressing enter)
to 0.6.6 which is just not the version installed locally when following
the instructions.


martinklepsch:
Thanks, good catch!!

PR #387, opened 2020-04-17
If namespaces has doctring but no vars, display it as well in sidebar

### LINKED ISSUE #21: Figure out what to do with empty namespaces
[Uruk](https://github.com/daveliepmann/uruk) contains some empty namespaces. Codox renders those as every other namespace, just empty. We should probably default to something similar. For now I removed them from the sidebar in https://github.com/martinklepsch/cljdoc/commit/f24ff7cd24210c49ee26258c8206e1727270388e

💡  **Good solution here:** https://github.com/cljdoc/cljdoc/issues/21#issuecomment-554664586

kloimhardt:
fixes #21, ready to merge into master
bundle.clj: in function has-defs, add or condition for :doc key
render/api.clj: format the two "No vars..." messages uniformly in italic blue

### DIFF
--- a/src/cljdoc/bundle.clj
+++ b/src/cljdoc/bundle.clj
@@ -9,7 +9,7 @@
   "Return entity-maps for all namespaces in the cache-bundle"
   [{:keys [version-entity] :as cache-bundle}]
   (let [nss-from-defs (set (map :namespace (:defs cache-bundle)))
-        nss-with-doc (set (map :name (filter :doc (:namespaces cache-bundle))))
+        nss-with-doc  (set (map :name (filter :doc (:namespaces cache-bundle))))
         has-defs?     (fn [{:keys [namespace]}]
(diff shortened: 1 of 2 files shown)


src/cljdoc/bundle.clj
```diff
@@ -9,8 +9,10 @@
   "Return entity-maps for all namespaces in the cache-bundle"
   [{:keys [version-entity] :as cache-bundle}]
   (let [nss-from-defs (set (map :namespace (:defs cache-bundle)))
-        has-defs?     (fn [ns-emap]
-                        (contains? nss-from-defs (:namespace ns-emap)))]
+        nss-with-doc (set (map :name (filter :doc (:namespaces cache-bundle))))
```
martinklepsch, line 12:
```suggestion
        nss-with-doc  (set (map :name (filter :doc (:namespaces cache-bundle))))
```

PR #389, opened 2020-04-20
Throw exception when no namespace is found

### LINKED ISSUE #299: False positive in detecting build failure
https://cljdoc.org/builds/13942 this page says the build failed but there's no exception in the logs and the CircleCI build ran through successfully.

```
INFO [2019-05-22 22:01:27,062] - cljdoc.analysis.service CircleCI: Polling build 5314
INFO [2019-05-22 22:01:28,567] - cljdoc.server.ingest Verifying cljdoc-edn contents against spec
INFO [2019-05-22 22:01:28,567] - cljdoc.server.ingest Importing API into database metosin/reitit-core 0.3.5
INFO [2019-05-22 22:01:28,574] - cljdoc.storage.sqlite-impl version-id 11290
```
The [build artifact](https://5297-162196335-gh.circle-artifacts.com/0/cljdoc-edn/metosin/reitit/0.3.5/cljdoc.edn) for the build contains no namespaces. There are no source files in the artefact so that is correct but maybe this is what hasn't been considered here.

kloimhardt:
Closes #299 

PR is technically ready to merge. The question remains on my side, whether throwing an exception is actually in the spirit and intention of issue #299.

### DIFF
--- a/src/cljdoc/server/api.clj
+++ b/src/cljdoc/server/api.clj
@@ -34,7 +34,8 @@
         (build-log/analysis-received! build-tracker build-id file-uri)
         (ingest/ingest-cljdoc-edn storage data)
         (build-log/api-imported! build-tracker build-id ns-count)
-        (build-log/completed! build-tracker build-id))
+        (build-log/completed! build-tracker build-id)
+        (when (zero? ns-count) (throw (Exception. "No namespaces found"))))
       (catch Exception e
         (log/errorf e "analysis job failed for project: %s, version: %s, build-id: %s" project version build-id)
         (build-log/failed! build-tracker build-id "analysis-job-failed")))))


src/cljdoc/server/api.clj
```diff
@@ -34,7 +34,8 @@
         (build-log/analysis-received! build-tracker build-id file-uri)
         (ingest/ingest-cljdoc-edn storage data)
         (build-log/api-imported! build-tracker build-id ns-count)
-        (build-log/completed! build-tracker build-id))
+        (build-log/completed! build-tracker build-id)
+        (when (zero? ns-count) (throw (Exception. "No namespaces found"))))
```
martinklepsch, line 38: I'm wondering if maybe `build-log/completed!` shouldn't be called when there's no namespaces. Maybe instead we just call `build-log/failed!` with the "No namespaces found" message?

kloimhardt: This is the current master behaviour:
Screenshot 2020-04-21 at 19 43 08

Here how this commit currently shows it:
Screenshot 2020-04-21 at 19 51 43

Here the result when I implement your suggestion (i.e. do not call completed!):
Screenshot 2020-04-21 at 20 09 41

Also possible is this (does neither call api-imported! nor completed!):
Screenshot 2020-04-21 at 20 17 01

I do not like the last version, not calling api-imported! seems odd to me. The current PR is the most prudent and least code-intrusive as everything that was called is still called and only a throw of the exception is added for the logging purposes.

martinklepsch: Ok sounds good! Thanks for taking the time to walk me through it, appreciate that!

kloimhardt: My pleasure, when the workpiece is already clamped, variations are easy :-)

martinklepsch: https://youtu.be/ShEez0JkOFw :)

martinklepsch:
Hm. I think I might have been misguided here. Reitit is an artifact that doesn't contain namespaces and that should be ok. 

For a new release it now looks as if the build failed when really it didn't https://cljdoc.org/builds/30671

I guess we should roll this back? 🤔

kloimhardt:
Please act at your discretion. With the PR #389 I tried to solve Issue #299, but never was entirely sure whether I succeed entirely with this one single line. Maybe it is wise to roll this back.

PR #390, opened 2020-04-30
Change the hardcoded config for re-frame

davidjameshumphreys:
A file was moved as part of day8/re-frame#528


martinklepsch:
Thank you! 🙏

PR #392, opened 2020-05-06
Add api/namespaces config for re-frame

superstructor:
day8/re-frame#216 has improved the core ns api docstrings.


martinklepsch:
@superstructor I'm happy to merge this but I'm also thinking maybe it would be best to just move this config into the re-frame project itself?

superstructor:
Good idea. I will make that change today 👍 as per [docs](https://github.com/cljdoc/cljdoc/blob/master/doc/userguide/for-library-authors.adoc#configuring-articles).

PR #393, opened 2020-05-06
Remove re-frame config

superstructor:
Moved to day8/re-frame in day8/re-frame#592

See [commit 18345ce](https://github.com/day8/re-frame/commit/18345ce09e046d72c5a3a42be6e09297460f6ebb)

Is there anything else that I need to do ?

Does this build off day8/re-frame master ?

### DIFF
--- a/modules/shared-utils/resources/hardcoded-projects-config.edn
+++ b/modules/shared-utils/resources/hardcoded-projects-config.edn
@@ -2,40 +2,6 @@
  "cider/cider-nrepl" {:cljdoc.api/platforms #{"clj"}}
  "com.rpl/specter" {:cljdoc.api/platforms #{"clj"}}
  "ventas-core/ventas-core" {:cljdoc.api/platforms #{"clj"}}
- "re-frame/re-frame" {:cljdoc.doc/tree [["Introduction" {:file "README.md"}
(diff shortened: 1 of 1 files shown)


martinklepsch:
> Is there anything else that I need to do ?
> Does this build off day8/re-frame master ?

cljdoc generally looks at your git repo at a specific revision. It extracts this revision from the  `pom.xml` `<scm>` section. [Re-frame 0.12.0 specifies](https://repo.clojars.org/re-frame/re-frame/0.12.0/re-frame-0.12.0.pom) `a531e6f99881886f821021c1f651094309c7c02c` as revision.

This allows cljdoc to reliably build docs for older versions without potentially mixing older API docs with newer articles ("prose docs").

So in order to have these new changes reflected on cljdoc you **will need to cut a new release.**

I understand that sometimes this can be a little annoying but I think it's a little bit of extra work that library maintainers have to do _some times_ in exchange for a very consistent and reliable experience for documentation consumers.

Let me know if this all makes sense. Also you can push a SNAPSHOT release to test on cljdoc (cljdoc always uses `master` as the revision for SNAPSHOT releases).

superstructor:
Ok thanks @martinklepsch 👍 I think its reasonable, no trouble to cut a release. We hope to do a release tomorrow so just hold until then and then will be good to go.

PR #399, opened 2020-05-31
add previously visited doc links on homepage

### LINKED ISSUE #282: Show list of last viewed projects on homepage
We track last viewed docs in localstorage and allow users to access those via `CMD-K`/`CTRL-K`:

screenshot 2019-03-01 at 11 36 46

This information could also be useful to show on the homepage to allow users to quickly navigate to documentation they recently used without typing/searching much.

Take a look at [`trackProjectOpened`](https://github.com/cljdoc/cljdoc/blob/01102870d1ecb05c577ef593621bd8c396c8baee/js/switcher.js#L21-L39) in `switcher.js` to understand how this information is stored and could be used.

kolharsam:
This is a WIP.

This should resolve #282

### DIFF
--- a/js/switcher.js
+++ b/js/switcher.js
@@ -2,6 +2,8 @@ import { Component, createElement, h } from "preact";
 import fuzzysort from "fuzzysort";
 import * as listSelect from "./listselect";
 
+const previouslyOpenedKey = "previouslyOpened";
+
 function isSameProject(p1, p2) {
   // I can't believe you have to do this
   return p1.group_id === p2.group_id && p1.artifact_id === p2.artifact_id; //&& p1.version == p2.version
@@ -23,7 +25,7 @@ function trackProjectOpened() {
   const project = parseCljdocURI(window.location.pathname);
   if (project) {
     var previouslyOpened =
-      JSON.parse(localStorage.getItem("previouslyOpened")) || [];
+      JSON.parse(localStorage.getItem(previouslyOpenedKey)) || [];
     // remove identical values
     previouslyOpened = previouslyOpened.filter(p => !isSameProject(p, project));
     previouslyOpened.push(project);
@@ -34,7 +36,7 @@ function trackProjectOpened() {
         previouslyOpened.length
       );
     }
-    localStorage.setItem("previouslyOpened", JSON.stringify(previouslyOpened));
+    localStorage.setItem(previouslyOpenedKey, JSON.stringify(previouslyOpened));
   }
 }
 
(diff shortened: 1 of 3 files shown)


martinklepsch:
>  I just need some help in figuring out to get the localstorage data into the clojure env.

Since this is all stored in the client we cannot render that information on the server and instead need to use JS to show it. The [`switcher.js`](https://github.com/cljdoc/cljdoc/blob/master/js/switcher.js) file is what shows up when you hit CMD+K and uses the same data from localstorage. 

I think essentially we could implement something like the `Switcher` component in that file bit minus the interactivity for changing results

https://github.com/cljdoc/cljdoc/blob/3ab0001040ae41f4ded4d42eb4286558d367e83f/js/switcher.js#L81-L225

#### To render a component into a div with a specific ID 

https://github.com/cljdoc/cljdoc/blob/3ab0001040ae41f4ded4d42eb4286558d367e83f/js/index.js#L18

I hope this makes sense and I'm sorry if this now means writing some JS when you came here to write some sweet Clojure 🤗

kolharsam:
@martinklepsch to render the required component where do I place the component.
Right, I have something like this in `home.clj`

```clojure
 [:div#recent-docs-visited]
 [:script {:src "/js/index.js"}]
```
and within  `index.js`
```javascript
render(h(RecentlyVisited), recently-visited-node);
```

The component is mounted but the search box is being rendered twice? Any clue as to how I can resolve this?

martinklepsch:
Does your `RecentlyVisited` component include the search input? It looks like you haven’t pushed the code that’s doing this so it’s a bit hard to tell.

kolharsam:
@martinklepsch I've added the commit's that you asked for. Here, if the `RecentlyVisited` component is rendered on index.js, then that's when I see 2 search boxes being rendered. Also, other imports apart from `index.js` don't work on `home.clj`.

martinklepsch:
Hi Sameer! To include the `recents.js` file you will need to explicitly require it in `index.js`, similar to how it's done for the switcher: 
https://github.com/cljdoc/cljdoc/blob/3ab0001040ae41f4ded4d42eb4286558d367e83f/js/index.js#L2

You may also want to require some components of `switcher.js` so that you can reuse them within `recents.js`. 

I would also suggest to move the line actually rendering the component into `index.js`, similar to this:
https://github.com/cljdoc/cljdoc/blob/3ab0001040ae41f4ded4d42eb4286558d367e83f/js/index.js#L18

I hope this helps!

martinklepsch:
Hey Sameer :wave: let me know if you had a chance to look at my pointers. If they weren't enough to help you with this I'm happy to provide more :)

kolharsam:
Hey Martin! I just haven't had the time in recent weeks to work on this. I shall definitely try this weekend and get back to you :)

martinklepsch:
Cool, no rush, just wanted to check in :)

kolharsam:
@martinklepsch I'm not able to find the time to build this out. Perhaps you could take over?

PR #409, opened 2020-08-11
Add links to prev and next article on doc page

### LINKED ISSUE #207: Navigation between articles
It would be useful to add navigation elements of some sort directly on articles, such as "Previous" and "Next" buttons at the bottom of each article to allow users continue reading without having to look back at the table of contents.

moonbrv:
This is my attempt to fix #207 

For some reason Icons does not work so I added just regular characters `<` and `>` instead arrows. Will be glad to hear any proposition on how it should look like

image

### DIFF
--- a/src/cljdoc/render.clj
+++ b/src/cljdoc/render.clj
@@ -28,6 +28,14 @@
          (layout/page {:title (str (util/clojars-id version-entity) " " (:version version-entity))
                        :description (layout/artifact-description version-entity (:description pom))}))))
 
+(defn positions
+  "return collection if indexes where element appears in collection"
+  [pred coll]
+  (keep-indexed (fn [idx x]
+                  (when (pred x)
+                    idx))
+                coll))
+
 (defmethod render :artifact/doc
   [_ route-params {:keys [cache-bundle pom last-build]}]
   (assert (:doc-slug-path route-params))
@@ -40,7 +48,13 @@
                    first)
         [doc-type contents] (doctree/entry->type-and-content doc-p)
         top-bar-component (layout/top-bar version-entity (-> cache-bundle :version :scm :url))
-        sidebar-contents (sidebar/sidebar-contents route-params cache-bundle last-build)]
+        sidebar-contents (sidebar/sidebar-contents route-params cache-bundle last-build)
+        flatten-article-tree (->> doc-tree
+                                  (remove #(contains? #{"Readme" "Changelog"} (:title %)))
+                                  doctree/flatten*)
+        article-idx (->> flatten-article-tree
+                         (positions #(= doc-slug-path (:slug-path (:attrs %))))
+                         first)]
     ;; If we can find an article for the provided `doc-slug-path` render that article,
     ;; if there's no article then the page should display a list of all child-pages
     (->> (if doc-type
@@ -56,7 +70,12 @@
                         :doc-html (fixref/fix (-> doc-p :attrs :cljdoc.doc/source-file)
                                               (rich-text/render-text [doc-type contents])
                                               {:scm (bundle/scm-info cache-bundle)
-                                               :uri-map (fixref/uri-mapping version-entity (doctree/flatten* doc-tree))})})})
+                                               :uri-map (fixref/uri-mapping version-entity (doctree/flatten* doc-tree))})
+                        :version-entity version-entity
+                        :prev-page (when (<= 0 (dec article-idx) (count flatten-article-tree))
+                                     (nth flatten-article-tree (dec article-idx)))
+                        :next-page (when (<= 0 (inc article-idx) (count flatten-article-tree))
+                                     (nth flatten-article-tree (inc article-idx)))})})
            (layout/layout
             {:top-bar top-bar-component
              :main-sidebar-contents sidebar-contents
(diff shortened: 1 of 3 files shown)


martinklepsch:
Hi @moonbrv, this looks great! I likely won't have time to give it a closer look until next week but if another @cljdoc/cljdoc-community member wants to merge this I'd be happy to see it go live. Feel free to bug folks in #cljdoc on Slack as well, most are very happy to help :)

src/cljdoc/render.clj
```diff
@@ -40,7 +48,17 @@
                    first)
         [doc-type contents] (doctree/entry->type-and-content doc-p)
         top-bar-component (layout/top-bar version-entity (-> cache-bundle :version :scm :url))
-        sidebar-contents (sidebar/sidebar-contents route-params cache-bundle last-build)]
+        sidebar-contents (sidebar/sidebar-contents route-params cache-bundle last-build)
+        flatten-article-tree (->> doc-tree
+                                  (remove #(contains? #{"Readme" "Changelog"} (:title %)))
+                                  doctree/flatten*)
+        article-idx (->> flatten-article-tree
```
danielcompton, line 55: Rather than working with indexes, you might be able to work directly with the article tree.

```clj
(partition 3 1 (concat [nil] (range 10) [nil]))
; => ((nil 0 1) (0 1 2) (1 2 3) (2 3 4) (3 4 5) (4 5 6) (5 6 7) (6 7 8) (7 8 9) (8 9 nil))
```

Replace `(range 10)` with `flatten-article-tree`. Then look for the seq where the middle element is the slug, then your `prev-page` and `next-page` are the first and the third element.

moonbrv: wow, that's good, thank you!

src/cljdoc/render/articles.clj
```diff
@@ -11,6 +11,18 @@
        (assoc version-entity :article-slug)
        (routes/url-for :artifact/doc :path-params)))
 
+(defn prev-next-navigation [{:keys [prev-page next-page version-entity]}]
+  [:div.flex.justify-between
+   (if prev-page
```
danielcompton, line 16: Should this be a when as well?

moonbrv: No, because if `prev-page` is `nil` we want to render empty`div` element, so `next-page` link always will be on the right side

moonbrv:
> Seems ok to me, tests would be good, but I'll let others decide if they should be blockers.

@danielcompton I will try cover functions for calculating prev and next page by tests, have not much experience with tests

moonbrv:
Added test. Not sure about quality of it... To make data for test autogenerated I had no change some specs and make separate specs for the test because I need `slug` in `attrs`. Existing spec makes `attrs` an optional key. May we change spec in for `entiy` in `cljdoc.doc-tree` to move `attrs` to `req-un` keys?

src/cljdoc/render.clj
```diff
@@ -40,7 +40,13 @@
                    first)
         [doc-type contents] (doctree/entry->type-and-content doc-p)
         top-bar-component (layout/top-bar version-entity (-> cache-bundle :version :scm :url))
-        sidebar-contents (sidebar/sidebar-contents route-params cache-bundle last-build)]
+        sidebar-contents (sidebar/sidebar-contents route-params cache-bundle last-build)
+        articles-block (doctree/get-neighbour-entries (remove #(contains? #{"Readme" "Changelog"}
```
avichalp, line 44: Thanks for the PR @moonbrv! 😃 

Should `remove` happen after flattening the `doc-tree`?  or "Readme" and "Changelog" are always expected in the root so it doesn't matter?

avichalp:
> For some reason Icons does not work so I added just regular characters < and > instead arrows.

@moonbrv: Could this be related to: https://github.com/reagent-project/reagent/issues/104#issuecomment-76828792 ?

src/cljdoc/render.clj
moonbrv: I decide to remove `Readme` and `Changelog` to have the same data as in the sidebar. I think for the navigating we need only articles (`doc-tree-with-rest` in the code snippet below`). I can change if need 

https://github.com/cljdoc/cljdoc/blob/d6c6fab0ef34dbb2bf3734ba0614466a401c20bb/src/cljdoc/render/sidebar.clj#L46-L50

moonbrv:
> > For some reason Icons does not work so I added just regular characters < and > instead arrows.
> 
> @moonbrv: Could this be related to: [reagent-project/reagent#104 (comment)](https://github.com/reagent-project/reagent/issues/104#issuecomment-76828792) ?

I noticed that service for icons that used in this project https://icon.now.sh/ does not work

moonbrv:
> > For some reason Icons does not work so I added just regular characters < and > instead arrows.
> 
> @moonbrv: Could this be related to: [reagent-project/reagent#104 (comment)](https://github.com/reagent-project/reagent/issues/104#issuecomment-76828792) ?

@avichalp  You gave a good Idea to user regular Unicode characters, thanks 👍

martinklepsch:
@moonbrv Thanks Dmitry and sorry about the broken icon service. Used for work for a long time now it doesn't anymore 🤷‍♂️  — I opened #410 to hopefully follow up on that. 

The code changes look great to me, if you think it's ready I'd be happy to get this merged. 🤗

moonbrv:
@martinklepsch I have nothing to add, you may merge

PR #415, opened 2020-09-28
Fuller local preview

lread:
This is a replacement for PR #340.

This PR supports a better local testing experience.

**RECAP**
I tried to tell the story of this change in the commits, but to recap:

**1) fully qualified cljdoc urls now converted to root relative**

This supports local testing - we don't want to reach out to cljdoc.org when running a local server for testing purposes. 

**2) links and images are now fixed up in docstrings**

We weren't doing this at all prior to this PR.
This change means that root relative references to images and links will now work from docstrings. 

**3) raw urls are now generic**

The url used to reference images as raw content is no longer tied to
`raw.githubusercontent.com` and now uses the more generic
`https://<git provider>/<owner>/<repo>/raw syntax`. This syntax works with
github, gitlab and bitbucket.

The sanitize query parameter required for raw svgs from github has been preserved for github.  

This does not specifically help the local testing experience but was something I addressed while fixing up urls.

**4) minor fixes to cli tools to support local testing.**

- cljdoc script can now be run from outside cljdoc dirs
- `ingest` command now has a bit clearer help and respects the `--rev` option
-  `offline-bundle` command is now clear about where zip file is created

**5) Docs updated**

ADR added that records my musings on running locally.

running locally docs have been revamped 
- try to empathize with reader roles more: the library author vs the cljdoc developer
- add an overview diagram
- use cljdoc-exerciser as example library
- include steps for full vs quick preview (this was as a result of ADR)
- complete walk through examples with commands when running from source vs from docker

minor updates library authors docs 

**TO POSSIBLY DEBATE**
Some things to question and possibly debate:

1. use of cljdoc-exerciser as the example app for running locally.  
I think is a reasonable choice as it also shows what formatting is possible with cljdoc. 
Another option might be a cljdoc hello world library that shows what is recommended rather than possible.  
But that could be a separate issue/PR if there is strong interest.  
Or cljdoc-exerciser could be updated to advise on what is recommended.

2. only root relative references from docstrings.
I chose to only support root relative references from docstrings because 1) we could calculate, but do not currently know the scm file path for namespaces 2) in my opinion it seems unlikely that docs/images will be mixed with source code.  So what to do when an SCM reference in a docstring is not root relative? I currently render the `href`/`src` as `#!cljdoc-error!ref-must-be-root-relative!`.  I feel that this at least gives some indication to the library author that their reference is invalid.  I propose that this could be improved/changed, if there is interest, in a separate PR.

3. that's all I could think of, my dear reviewer, maybe you see something else I did not? 🙂


lread:
Note that this PR closes #325

doc/adr/0020-support-links-from-docs-to-SCM-when-testing-locally.md
```diff
@@ -0,0 +1,125 @@
...
+
+### Background
+
+Cljdoc imports documentation from clojars artifacts and SCM (Source Code Management) repositories. 
+The SCM is a git repository and usually one hosted on GitHub.
+
+Documentation is imported to cljdoc from 2 distinct sources:
+
+1. API docstrings from sources in a clojars artifact 
+2. Stand-alone articles such as `README.md` and documentation under `doc/** from an SCM repository.
```
martinklepsch, line 19:
```suggestion
2. Stand-alone articles such as `README.md` and documentation under `doc/**` from an SCM repository.
```

doc/running-cljdoc-locally.adoc
```diff
@@ -0,0 +1,461 @@
...
+
+* `project/scm/url` to be overridden via `--git`
+* `project/scm/tag` to be overridden via `--rev`
+
+For `--git` you can specify either:
+
+* Your library's root dir on your local filesystem for a quick preview. +
+In a quick preview, direct references to GitHub won't work.
+* The https GitHub project URL for a full preview. +
+In a full preview, direct references to GitHub will work.
```
martinklepsch, line 298: I find the wording here a little confusing. I think by direct references to GitHub we mean stuff like 

```
image::running-local.svg[running locally overview]
```

but when reading "direct references to GitHub" I'm more thinking of something like `https://github.com/...` — which should always work, right? 

Using a table here could be nice to make this stand out a bit more and contrast the different options.

doc/running-cljdoc-locally.adoc
```diff
@@ -0,0 +1,461 @@
...
+
+From {example-project-name}'s root dir:
+
+.cmd for cljdoc source - quick preview
+[%collapsible]
```
martinklepsch, line 303: These code blocks are the core of this guide and in my personal experience I often would just look at the snippets, see if they make sense and run them without actually reading the rest. With that in mind I have two suggestions:

- Don't hide them by default
- Use a table to contrast Docker/Local approach 🐳  The code snippets seem narrow enough that a table could work (at least on non-mobile screens).

doc/running-cljdoc-locally.adoc
```diff
@@ -0,0 +1,461 @@
...
+
+TIP: Your installation might require you to prefix docker commands with `sudo`, adapt as necessary.
+
+=== Steps
+
+Repeat steps 1 through 4 as needed (see overview diagram).
+Confusion can usually be resolved by asking:
+
+* Did I remember to commit/push?
+* Did I remember to publish locally?
```
martinklepsch, line 215: I think this might be a good candidate for an admonition `TIP` block now that this is an asciidoc file :) https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#admon-bl

modules/cli/src/cljdoc/cli.clj
```diff
@@ -58,14 +59,19 @@
                                 "You may specify full paths to those files using the --jar and --pom options"
                                 "(in this case --version may override version from pom)."
                                 ""
-                                "To test how a Git repository gets incorporated without pushing new release,"
-                                "pass the --git option and use --rev to specify a revision to use (e.g. master)."]
+                                "To test how a Git repository is ingested before pushing new release,"
+                                "use the --git and --rev options to specify a revision to use (e.g. master)."]
```
martinklepsch, line 63: 👍

script/cljdoc
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 
-cd $(git rev-parse --show-toplevel)
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+cd "$SCRIPT_DIR/.." || exit
```
martinklepsch, line 4: This is so that the script also works when not ran inside a Git repo?

src/cljdoc/render/api.clj
```diff
@@ -51,24 +53,24 @@
     (str (ns-link-fn (if ns (util/replant-ns current-ns ns) current-ns))
          (when var (str "#" var)))))
 
-(defn render-doc [mp render-wiki-link]
+(defn render-doc [mp render-wiki-link fix-opts]
```
martinklepsch, line 56: Given that we now pass down `render-wiki-link` and `fix-opts` all the way down to render the `doc-str` maybe it would make sense to instead pass down a function that takes `doc-str` and returns html? Might require some renaming of `docstring->html` but seems like it might be worth it? 

Ideally the "constructor" for that function would be easy to use within REPL sessions.

martinklepsch: I'd also be perfectly ok with merging this first and creating an issue to follow up with a small refactor.

src/cljdoc/server/ingest.clj
```diff
@@ -37,13 +37,16 @@
...
 (defn ingest-git!
   "Analyze the git repository `repo` and store the result in `storage`"
-  [storage {:keys [project version scm-url local-scm pom-revision] :as _repo}]
+  [storage {:keys [project version scm-url pom-revision requested-revision] :as _repo}]
   {:pre [(string? scm-url)]}
-  (let [git-analysis (ana-git/analyze-git-repo project version (or local-scm scm-url) pom-revision)]
+  (let [scm-rev (or requested-revision pom-revision)
+        git-analysis (cond-> (ana-git/analyze-git-repo project version scm-url scm-rev)
+                       ;; The git tag expressing the project version is not relevant if git revision specifically requested
+                       requested-revision (update-in [:scm] dissoc :tag))]
```
martinklepsch, line 45: Could passing the revision twice here somehow be solved at the call site? It seems that the only reason we need `requested-revision` is to remove the `[:scm :tag]` field. Maybe we could use `(not= scm-rev (get ana-result [:scm :tag])` as a condition to remove it instead and this way avoid the extra arg?

src/cljdoc/util/fixref.clj
```diff
@@ -35,77 +109,60 @@
...
-  ;; (def hs html-str)
-  ;; (def fo fix-opts)
-  (let [doc     (Jsoup/parse html-str)
-        scm-rev (or (:name (:tag scm))
-                    (:commit scm))]
-    (doseq [broken-link (->> (.select doc "a")
-                             (map #(.attributes %))
-                             (remove #(absolute-uri? (.get % "href")))
-                             (remove #(anchor-uri? (.get % "href"))))]
+  "Rewrite references in html produced from rendering markdown.
```
martinklepsch, line 113:
```suggestion
  "Rewrite references in HTML produced from rendering markdown.
```
😄

martinklepsch (commented):
Hi @lread! Thanks for getting back to this 🤗  I think the decision you made with regards to requiring people to push to their SCM makes perfect sense and I love all the updates to the documentation and that graph illustrating it all 😍 

There are a few changes to the documentation formatting that I think would be great to get in as well as some ideas for refactors but overall I think this looks really solid! 👏  🙇

doc/running-cljdoc-locally.adoc
lread: Agreed, will work on this! Tx!

doc/running-cljdoc-locally.adoc
lread: Sure, makes sense, will do!

script/cljdoc
lread: Yes, that's correct.

It is convenient to run `ingest` from your library's git repo dir, and this change allows for that:

`../cljdoc/script/cljdoc ingest` ...

src/cljdoc/render/api.clj
lread: I like that idea, I'll take a look.

src/cljdoc/server/ingest.clj
lread: Probably.  I'll take a look.

doc/running-cljdoc-locally.adoc
lread: This one was tricky for me and I went through more variations than I'd like to admit. 🙂 

I felt the hidden code blocks allowed the headings and descriptive text (and therefore outline) to be visible on the page.  This variation felt less overwhelming to me and I got less lost on the page.  

But these are subjective things. 

Let me know if you feel strongly about this and I can experiment more.

lread: I do think I see your point now. I think I was just getting tired of editing this doc.  I'll give it a shot.

src/cljdoc/server/ingest.clj
lread: I can easily remove the extra arg but I'll keep the removal logic the same.

`:scm` -> `:tag` -> `:name` is preferred over `:scm` -> `:commit` [at url fixup time](https://github.com/cljdoc/cljdoc/blob/966b30ef8833143c97a7d700fd6e8ac9b265c5fc/src/cljdoc/util/fixref.clj#L77).
`:scm` -> `:tag` -> `:name` is the tag representing the project version, if it exists (ex "v1.2.3" or "1.2.3"), as far as I understand it.

I had a todo to check if this current preference logic is appropriate/consistent/documented but did not get to it.  My guess the intent was to provide a nicer fixed up URL (version looks nicer than git SHA), but I don't know.

lread: Oh wait... I see why I did what I did: `ingest-git!` analyzes AND stores, so I'm thinking, the way things are structured now... that extra arg is necessary to make the decision to turf `:scm` -> `:tag` after analysis and before storage.

src/cljdoc/render/api.clj
lread: You know what? Let's try to get this to master while we have momentum. 🙂  (A follow up issue makes sense).

martinklepsch: Sounds good to me! 🙂 🚀

doc/running-cljdoc-locally.adoc
martinklepsch: Thanks for persevering 🙇

martinklepsch (approved):
SHIP IT 🚢  🙌 😄

PR #429, opened 2020-11-15
Prevent default behavior for Ctrl-K

renatoalencar:
On Firefox and Chrome (the ones I tested), when you press `Ctrl+K`
the browser's navigation bar gets focused and adds a question mark
for searching, which prevents the user from using this as an actual
feature. Calling `preventDefault` on the event, makes the cljdoc
search box to get focused instead.

**Reference**:

  * Chrome: https://support.google.com/chrome/answer/157179
  * Firefox: https://support.mozilla.org/en-US/kb/keyboard-shortcuts-perform-firefox-tasks-quickly#w_search

**Note**: I don't know if this could be the better option, if you're trying not to obstruct the user's usual interaction with the browser. But, at least, letting it here as a note.

### DIFF
--- a/js/switcher.js
+++ b/js/switcher.js
@@ -105,6 +105,7 @@ class Switcher extends Component {
 
     if (e.which === 75) {
       if ((isMac && e.metaKey) || (!isMac && e.ctrlKey)) {
+        e.preventDefault();
         switcherShortcut = true;
       }
     }


martinklepsch:
Interesting, thanks for looking into this. The intention was to mirror the interface many people may be familiar with from Slack and other apps. It seems that on Windows (which you are using I believe?) there are additional shortcuts mapped to <kbd>ctrl</kbd> + <kbd>k</kbd>.

If you happen to use Slack, how does this work in Chrome on Windows?  Probably I'd suggest to just follow their lead until someone drops by & complains 😇

renatoalencar:
> Interesting, thanks for looking into this. The intention was to mirror the interface many people may be familiar with from Slack and other apps. It seems that on Windows (which you are using I believe?) there are additional shortcuts mapped to ctrl + k.

I'm using Ubuntu actually, and Firefox.

> If you happen to use Slack, how does this work in Chrome on Windows? Probably I'd suggest to just follow their lead until someone drops by & complains innocent

It works just fine, it doesn't focus on the browser's navigation bar, and I believe that could be a good options. If cljdoc users tend to be Slack users as well, I believe they would be fine.

martinklepsch (approved):

martinklepsch:
Well, then it's going live :)

PR #440, opened 2021-03-21
Render AsciiDoc document titles

### LINKED ISSUE #439: AsciiDoc document titles are not rendered
If we look at cljdoc-exerciser, we can see that adoc file titles are not rendered:

image

lread:
Fixes #439


martinklepsch:
Thank you Lee 🙏

PR #442, opened 2021-03-24
Add MathJax Support for AsciiDoc STEM formulas

### LINKED ISSUE #434: ability to inject mathjax onto cljdoc pages?
In https://github.com/sicmutils/sicmutils we have a lot of mathematical functions that would benefit from the ability to render LaTeX on our article pages (or in docstrings, now that I think of it!!)

I don't think we need to do anything special, other than all the user to opt some script block like this in to the footer of the page:

```
script>
window.MathJax = {
  loader: {load: ['[tex]/physics']},
  tex: {
    tags: 'ams',
    processEscapes: true,
    inlineMath: {'[+]': [['$', '$']]},
  }
};
</script>
<script type="text/javascript" id="MathJax-script" async
  src="https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js">
</script>
```

This would let forms like $\sin \phi$ render as TeX.

lread:
Notes to reviewers (I think another set of eyeballs on this one is a good idea):
- Online support seemed pretty straightforward to me, the only note I have is:
  - we do not bring in MathJax assets (from CDN) unless an AsciiDoc article uses STEM.
- Downloaded doc support was a bit "interesting":
  - brought in a manually curated subset of the girthy MathJax for offline viewing, adds about ~600kb to the downloaded doc zip file (for comparison the STEM-less downloaded rewrite-clj docs zip file weighs ~190kb).
  - only adding MathJax to zip when it is used in a library's docs
  - disabled the MathJax menu for offline docs to help slim/simplify things a bit more

As always questions & feedback welcome.

Closes #434


martinklepsch:
This looks really promising!

> brought in a manually curated subset of the girthy MathJax for offline viewing, adds about ~600kb to the downloaded doc zip file (for comparison the STEM-less downloaded rewrite-clj docs zip file weighs ~190kb).

Do you think there is a way that we could integrate those files without inlining them into our repository? I think some other assets are downloaded from CDNs on the fly when building the offline bundle, ideally we'd do something like that for these assets as well.

lread:
> Do you think there is a way that we could integrate those files without inlining them into our repository? I think some other assets are downloaded from CDNs on the fly when building the offline bundle, ideally we'd do something like that for these assets as well.

Thanks kindly for the review @martinklepsch!

Yeah, I suppose you are right, they should be all up there on CDN for fetching.
I have hand-picked ~70 files from MathJax for offline support.
MathJax v2.x doesn't seem to have a single minified js, so it will likely be ~70 downloads.
I'll just have to figure out what form/name these take, but I expect that should not be too hard (I'll letcha know!).
If we decide downloading these files every time they are needed is not reliable/performant, we can always introduce some form of caching.

I expect the above should work out fine, but if it does not, a compromise would be to store mathjax offline support in a zip in the cljdoc repo - so as to include one 3rd party thing and not pepper the repo with many little 3rd party things (which I expect is your main concern here?).

lread:
@martinklepsch I've given downloading assets from CDN a whirl in 79bd4d57c9ef6120244bcf10e161a2650c534c72, when you have some time and interest would like to hear what you think.

My concerns are:
- parsing asset-deps.edn for each page render might be excessive, could memoize I suppose.
- re-downloading static assets from CDN for offline zip files might also be excessive, but... I don't think this feature is use too much?
- building a zip file in memory, should we really be doing that?  Or do we have plenty to spare?

Whaddya think?

martinklepsch (approved):

lread:
Thanks for addressing my concerns @martinklepsch, your responses all make sense to me.

I shall make it live!

PR #443, opened 2021-03-26
Tweak calls to badge service

### LINKED ISSUE #422: Consider options for badge service
Heyo, I notice that cljdoc badges are currently not rendering.

image

If I try to "view image" from Firefox, I get:

```
XML Parsing Error: syntax error
Location: https://camo.githubusercontent.com/84d27b26e179df25d064b47751b7aac556cbda2b/68747470733a2f2f636c6a646f632e6f72672f62616467652f6c726561642f636c6a646f632d657865726369736572
Line Number 1, Column 1:
Badge service error for URL https://badgen.net/badge/cljdoc/1.0.53/blue
^
```

Assuming we are not doing anything wrong, it might be interesting to accept that badge services ae not entirely reliable and consider a fallback badge service. 
If we encounter a non success from our primary, we could fallback to our secondary.

Maybe shields.io would be a good fallback?

The badges aren't identical, but they are close enough, I think.

Badgen Live Sample: 

Shields.io Live Sample:

lread:
Because badgen.net will often (always?) return an error on the first
call to a badge it does not have in its cache, we introduce 1 retry
on failure after a 300ms sleep.

Also we now return a 503 to the client if we can't manage to get
a sucessful return from badgen.net. We were previously returning 200
on failure with a request to cache for 30 minutes, so a browser refresh
would not always retry to fetch a failed badge.

Fixes #422

### DIFF
--- a/src/cljdoc/server/pedestal.clj
+++ b/src/cljdoc/server/pedestal.clj
@@ -34,7 +34,7 @@
             [cljdoc.util.sentry :as sentry]
             [clojure.tools.logging :as log]
             [clojure.string :as string]
-            [clj-http.lite.client]
+            [clj-http.lite.client :as http-client]
             [co.deps.ring-etag-middleware :as etag]
             [integrant.core :as ig]
             [io.pedestal.http :as http]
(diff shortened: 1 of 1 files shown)


lread:
@martinklepsch this seems to work well in testing from my dev box. 
How does logging look, does it work for you, and give you the info you need/want?

martinklepsch (approved):

PR #462, opened 2021-08-06
Add retries to maven fetches.

corasaurus-hex:
The artifact indexer was failing for me every time it ran. The maven
servers kept giving us back 429 responses which mean we should back
off on our request rate a little.

So for the implementation I went to add retries with backoff and
discovered that we're using io/reader with a url to fetch from maven.
Since there's no way to get a response map and handle this another way
we need to catch the exception and retry the operation.

There are libraries we could pull in to catch exceptions and retry
operations but they're all a little overcomplicated in their own way.
Given how simple of a need we have it seemed a good idea to write it
ourselves. If our needs grow then it's probably a good idea to pull
in a library instead.

### DIFF
--- a/src/cljdoc/server/search/artifact_indexer.clj
+++ b/src/cljdoc/server/search/artifact_indexer.clj
@@ -1,6 +1,7 @@
 (ns cljdoc.server.search.artifact-indexer
   (:require
    [cljdoc.spec :as cljdoc-spec]
+   [cljdoc.util.with-retries :refer [with-retries]]
    [clojure.edn :as edn]
    [clojure.spec.alpha :as s]
    [clojure.string :as str]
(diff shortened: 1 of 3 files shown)


corasaurus-hex:
Oh, and using a trampoline was inspired by https://github.com/joegallo/robert-bruce/blob/master/src/robert/bruce.clj since you can't recur from inside of a try/catch.

corasaurus-hex:
I think we could just replace the `with-open` calls with calls to `clj-http-lite` like: 

```
(client/get "http://site.com/bigrequest.html" {:as :stream})
```

We'd get the same streaming response body and not have to deal with `with-retries`...

martinklepsch:
I'll try to give this a more thorough review next week but two quick thoughts: 

- Let's totally use `clj-http-lite`, feels strange to just use `io/reader` to be honest :D 
- I think it would be ok to use `robert-bruce`. It seems well tested and does what we need in a small package. 
- I'm a fan of not using `:refer` so that you can always see where a var is coming from.

PR #463, opened 2021-08-06
Fix bb.edn so clojure-lsp doesn't crash.

corasaurus-hex:
The most recently released version clojure-lsp has an issue with paths
being symbols even though babashka allows them. See link.

https://github.com/clojure-lsp/clojure-lsp/issues/507


lread:
Oh that was me that did that! Thanks for the fix.

lread:
@corasaurus-hex, as a past contributor, you should have the power to merge this yourself! Please feel free to do so. 🚀

corasaurus-hex:
Sounds good!

PR #464, opened 2021-08-06
Retry maven fetches with clj-http-lite.

corasaurus-hex:
The artifact indexer was failing for me every time it ran. The maven
servers kept giving us back 429 responses which mean we should back
off on our request rate a little.

So for the implementation I went to add retries with backoff and
discovered that we're using with-open with io/reader on a url to pull
docs from maven. This is less than ideal and we have clj-http-lite
already in the project so it makes sense to switch to that.

This change moves to clj-http-lite and adds robert-bruce for retrying
with backoff on exceptions when requests fail. Now download and
indexing doesn't fall over!

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -43,6 +43,7 @@
         org.apache.lucene/lucene-core {:mvn/version "8.1.1"}
         org.apache.lucene/lucene-analyzers-common {:mvn/version "8.1.1"}
         org.apache.lucene/lucene-queryparser {:mvn/version "8.1.1"}
+        robert/bruce {:mvn/version "0.8.0"}
 
         ;; These deps are dependencies of shared-utils and are
         ;; duplicated in the global modules/shared-utils/deps.edn,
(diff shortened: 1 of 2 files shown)


lread:
Thanks! Nice clean, focused, and easy to follow!

When I searched for what rate limiting the maven servers actually implement, I did not find anything.  So, this retry with back-off strategy makes sense to me. If they change their rules, we can always tweak our back-off.

In your testing I'm guessing you saw a 429 followed by a successful retry?

(Reviewing also helped me to understand what cljdoc is actually fetching from maven.  Surprisingly little!)

Unless @martinklepsch has any thoughts/concerns, this looks good to merge to me.

corasaurus-hex:
Yep! We get 429s followed by retries. I could add the logging back in? It would make the logs really chatty and 429s are expected, so I vote not to.

corasaurus-hex:
I was looking again and realized I had wrapped fetch-body in a retry when fetch-body already contains a retry. I removed that, simplifying this further.

lread:
> Yep! We get 429s followed by retries. I could add the logging back in? It would make the logs really chatty and 429s are expected, so I vote not to.

I agree with you.

> I was looking again and realized I had wrapped fetch-body in a retry when fetch-body already contains a retry. I removed that, simplifying this further.

A good test to see if reviewers are up to snuff. And well... 🙂

PR #467, opened 2021-08-15
Switch to TypeScript

corasaurus-hex:
I'm not familiar with ClojureScript front-end frameworks but I have
quite a bit of experience with TypeScript. It's not a huge change to
switch to TypeScript, especially since we use Parcel, so what this
gives us is some more certainty when adding incremental features
now until we have a better direction on the front-end later.

If you use VSCode for the front-end you'll find some really, really
nice hints and warnings when things go awry and a whole lot of
confidence when things are going smoothly.

I haven't done any in-browser testing but the type-checker is
happy and the builds work which goes a long, long way towards
confidence that this is working now.

Also, bonus: I found a few bugs by switching over.


lread:
Interesting! I don't remember having ever touched this part of the code base (if I have it wasn't in any big way).

Without ever studying it at all, I did sometimes wonder why cljdoc JavaScript wasn't written ClojureScript.

I personally don't have any strong opinions on switching to TypeScript and defer to our cljdoc leader, @martinklepsch.

The con, I suppose, is that anybody wanting to work on this area of the code would have to understand TypeScript.
Perhaps the benefits of TypeScript outweigh this cost.
I don't personally know TypeScript but glancing at your diff makes me think I could certainly get by.

corasaurus-hex:
Also to clarify, I didn't really want to add the in-docset search frontend until I had more confidence in how everything was working and that everything was working correctly, and this gets me a long way towards that.

lread:
> Also to clarify, I didn't really want to add the in-docset search frontend until I had more confidence in how everything was working and that everything was working correctly, and this gets me a long way towards that.

That certainly goes in the "pro" column!

corasaurus-hex:
Also probably worth noting that I switched the remaining `h` calls to JSX and updated deps and fixed some outdated browser API usage.

corasaurus-hex:
I believe everything is working correctly now, with strict type checking so we can be sure we're handling all the odd cases.

PR #470, opened 2021-09-14
Add :url option for cljdoc.edn sidebar links #433

### LINKED ISSUE #433: Add :url option for cljdoc.edn sidebar links
A note from @martinklepsch in Slack:

> I would be open to a change to allow a :url key and showing that in the sidebar with adequate markings

This would allow forms like:

```clojure
{:cljdoc.doc/tree
 [["Introduction" {:file "doc/intro.adoc"}]
  ["#sicmutils on Slack" {:url "https://clojurians.slack.com/blah"}]]}
```

rollacaster:
Hi, here's a first draft to close #433.

I did not add any tests since simiplar parts of cljdoc are not tested and adding those tests is probably a separate issue.

Some open questions I have:
- The key `:slug` is not qualified key in a `doctree` while all other keys are qualified. I used a qualified keyword `:cljdoc.doc/external-url` and I am not sure if that's ok?
- Should I throw an exception if the URL is invalid? (similar to the file handling)
- If `file` and `url` are specified in a `:cljdoc.doc-tree/hiccup-entry` the `url` will take precedence, is that ok or should I forbid this cases in general via spec?


corasaurus-hex:
Ahhh, yep, looks like it would need to be added to the [for-library-authors.adoc](https://github.com/cljdoc/cljdoc/blob/master/doc/userguide/for-library-authors.adoc) (thanks @lread)

corasaurus-hex:
It would also be good to mark these as external links with an icon like external link `https://microicon-clone.vercel.app/external/12`

corasaurus-hex:
Oh, and one last thing! The external links should have `target="_blank"` so that they open in a new tab/window.

rollacaster:
Thanks a lot for the review!

- I tried to improve the spec of `cljdoc.edn` and used `map-of` to forbid invalid configs. I am not sure if my approach is good. I am quite new to spec. As far as I understood one of the core ideas is to allow every map to be extendable. My solution is working against that so I am happy to receive some guidance if this is OK or not.
- I used `java.net.URI.` to spec the URL. It's allowing quite a lot so it's still possible to create invalid URLs but it's better than before. Another option would be to actually do a request to check the URL. I think `process-toc` is only called during the analysis so this might be OK, what do you thing about that? Or is a better idea to do this in a separate PR?

src/cljdoc/doc_tree.clj
```diff
@@ -56,13 +58,15 @@
...
+(spec/def ::file-attrs
+  (spec/map-of #{:file} ::file))
+(spec/def ::url-attrs
+  (spec/map-of #{:url} :cljdoc.doc/external-url))
 
 (spec/def ::hiccup-entry
   (spec/spec
    (spec/cat :title ::title
-             :attrs (spec/? ::hiccup-attrs)
+             :attrs (spec/? (spec/or :file-attrs ::file-attrs :url-attrs ::url-attrs))
```
corasaurus-hex, line 69: I think we'd be better off going with this on line 61:

```clojure
(spec/def ::hiccup-attrs
  (spec/and (spec/keys :opt-un [::file ::url])
            #(not (and (:file %) (:url %)))))
```

And then on this line go back to using:

```clojure
(spec/? ::hiccup-attrs)
```

Then we could revert some of the changes below which would reduce a _lot_ of churn. That seems cleanest to me but I'm interested if there are other reasons this is better than I'm not seeing just yet.

src/cljdoc/doc_tree.clj
```diff
@@ -30,6 +30,7 @@
 (spec/def :cljdoc.doc/type #{:cljdoc/markdown :cljdoc/asciidoc})
 (spec/def :cljdoc/asciidoc string?)
 (spec/def :cljdoc.doc/contributors (spec/coll-of string?))
+(spec/def :cljdoc.doc/external-url (spec/and ::ne-string #(try (java.net.URI. %) true (catch Exception _ false))))
```
corasaurus-hex, line 33: You're right that this accepts a lot of different URLs that aren't particularly helpful. I think we might want to do further checking beyond this. Maybe we should just [check for a scheme of http/https](https://docs.oracle.com/javase/8/docs/api/java/net/URI.html#getScheme--) (case insensitive) and [a hostname that's not empty](https://docs.oracle.com/javase/8/docs/api/java/net/URI.html#getHost--)? Everything else should be optional, really? That seems like a really simple change, it could be thrown in a function.

Also, this should only catch `URISyntaxException`, just so that we're not accidentally swallowing other exceptions.

src/cljdoc/render/articles.clj
```diff
@@ -33,12 +33,17 @@
    (when (seq doc-bundle)
      (->> doc-bundle
           (map (fn [doc-page]
-                 (let [slug-path (-> doc-page :attrs :slug-path)]
+                 (let [slug-path (-> doc-page :attrs :slug-path)
+                       external-url (-> doc-page :attrs :cljdoc.doc/external-url)]
                    [:li
                     {:class (when (seq (:children doc-page)) "mv2")}
+                    (when external-url
+                      [:img.v-mid.mr1 {:src "https://microicon-clone.vercel.app/external/12"}])
```
corasaurus-hex, line 41: I'm not sure what this looks like but I'm guessing this puts the image to the left of the link. I think we'd want it on the right of the link, though? I guess I'd have to see it. I can futz with the styling later if you don't want to, really.

src/cljdoc/render/articles.clj
```diff
@@ -33,12 +33,17 @@
...
                     [:a.link.blue.hover-dark-blue.dib.pv1
                      {:style {:word-wrap "break-word"}
-                      :href  (doc-link version-entity slug-path)
+                      :target "_blank"
+                      :rel "noopener"
```
corasaurus-hex, line 45: Oh, shoot, I wasn't very explicit here. We only want `:target "_blank" :rel "noopener"` when there's an external-url.

test/cljdoc/doc_tree_test.clj
```diff
@@ -22,12 +22,15 @@
                                :cljdoc.doc/type :cljdoc/asciidoc
                                :cljdoc.doc/contributors ["A" "B" "C"]
                                :cljdoc/asciidoc "nested.adoc"
-                               :slug "nested"}}]}]
+                               :slug "nested"}}]}
+          {:title "Community-Page",
+           :attrs {:cljdoc.doc/external-url "http://my-community.com", :slug "community-page"}}]
```
corasaurus-hex, line 27: 🤩 🤩 🤩 🤩 🤩 🤩 🤩

corasaurus-hex (changes requested):
This is shaping up really well! I hate throwing back more changes again, but I think it needs them to really feel complete, if you know what I mean? Thank you **so** much for your contribution!! (💜)

src/cljdoc/doc_tree.clj
corasaurus-hex: Maybe it wouldn't reduce churn at all and it still makes sense to break the file handling out into its own function? In which case this is fine as-is. I'll leave it up to you 🤷🏻‍♀️ .

src/cljdoc/doc_tree.clj
martinklepsch: FWIW I think in the interest of unblocking this contribution I'd be totally fine with this approach. It might not be perfect but it's certainly an improvement. In a tradeoff between making the contribution process easy and fun and perfection I usually tend to chose the former — which may or may not be a good decision down the line 😂

martinklepsch (commented):
Thanks @rollacaster for the contribution + @corasaurus-hex for stepping up and reviewing 🙌

test/cljdoc/doc_tree_test.clj
corasaurus-hex: 🐌 🐌 🐌 🐌

src/cljdoc/render/articles.clj
corasaurus-hex: Ok I did a bunch of research to see what the deal was here and I'm on-board with this now, just remove the `:target` and `:rel` imo. Sorry for the runaround!

src/cljdoc/doc_tree.clj
rollacaster: Since there's the other issue with `slug` down below, I think it's easier to keep the processing separated although it's a bigger change. 

Otherwise I'd need one more case to make sure the `slug` is not added which seems to be more coupling than necessary.

src/cljdoc/doc_tree.clj
rollacaster: You definitely don't have to worry about the all the comments. I want to become a better Clojure dev and I appreciate all the feedback I get here.

rollacaster:
I fixed the issues mentioned above. But I found one more issue:

- How to handle the next/previous links on the bottom of each doc page? In the current implementation I ignore any external links and the next links always points to the next doc page. I am not sure if this is a good idea since it might be confusing. The alternative of showing them as a next link might be confusing as well. You would leave `cljdoc` after a click on the next page.

After working for some time I am also not sure if it'a good idea to add this feature. It might do more harm than good:

- The headline of any external link is `ARTICLES` while most external links probably won't be articles.
- The solution of the prev/next page as described above could be confusing
- Most of the code assumes that a page of `cljdoc` has a `slug`. External links are an exception of that. This makes the code more complex.

Possible alternatives to the current implementation are:
- Add external links to the content of your documentation. (which means just won't fixing the issue)
- Add external links between `ARTICLES` and `NAMESPACES` on the sidebar. But for this idea the structure of `cljdoc.edn` needs to be changed. (a separate tree for external links)
- Add external links between `Readme/Changelog` and `ARTICLES` on the sidebar. But for this idea the structure of `cljdoc.edn` needs to be changed. (a separate tree for external links)

What do you think about that? Would you still add the feature as it is, change it or revert it?

doc/userguide/for-library-authors.adoc
```diff
@@ -116,10 +117,11 @@ Which will result in the following hierarchy being shown in your docs:
 ├── Readme
 ├── Getting Started
 │   └── Installation
-└── Configuration
+├── Configuration
+└── #my-lib on Slack
 ----
 
-IMPORTANT: The resulting URLs for those articles will be based on the title provided in the `cljdoc.edn` file and _not_ on the filename.
+IMPORTANT: The resulting URLs for those articles will be based on the title provided in the `cljdoc.edn` file and _not_ on the filename. External links will open in a new tab.
```
martinklepsch, line 124: This is no longer correct right?

src/cljdoc/doc_tree.clj
```diff
@@ -83,35 +95,36 @@
 (defmethod filepath->type "md" [_] :cljdoc/markdown)
 (defmethod filepath->type "adoc" [_] :cljdoc/asciidoc)
 
-(defn- process-toc-entry
-  [{:keys [slurp-fn get-contributors] :as fns}
-   {:keys [title attrs children]}]
-  {:pre [(string? title) (fn? slurp-fn) (fn? get-contributors)]}
+(defn- process-file-attrs [{:keys [slurp-fn get-contributors]}
+                           title {:keys [file slug]}]
```
martinklepsch, line 99: From a formatting perspective this might be easier readable as 
```suggestion
(defn- process-file-attrs 
  [{:keys [slurp-fn get-contributors]}
   title
   {:keys [file slug]}]
```
Generally when I have to add a linebreak between arguments I add one between every argument for consistency (although you'll probably find exceptions within this code base where I didn't do that :))

src/cljdoc/doc_tree.clj
```diff
@@ -152,16 +165,18 @@
   "For various purposes it is useful to know the path to a given document
   in it's doc tree. This function adds a field in [:attrs :slug-path] that
   contains the slug of the current document preceded by all slugs of documents
-  above it in the tree.
+  above it in the tree. Entries without a slug are ignored.
```
martinklepsch, line 168:
```suggestion
  above it in the tree. External URLs that don't have slugs are ignored.
```
Both fine, just felt like that tells the person reading this why there might be entries without slugs.

rollacaster:
> I'm not sure what you are referring to when you say "headline", could you explain?

I mean this gray section headline ARTICLES:
image

I assume that external links will rarely be articles. The headline would be incorrect for them. So maybe a separate section would make sense:

image

but to do this `cljdoc.edn` would need to be changes to sth. like:

```
{:cljdoc.doc/tree [["Readme" {:file "README.md"}]
                   ["Document tests" {:file "doc/overview.adoc"}
                    ["AsciiDoctor features" {:file "doc/tests/adoc-features.adoc"}]
                    ["CommonMark features" {:file "doc/tests/md-features.md"}]
                    ["AsciiDoctor User Manual" {:file "doc/tests/asciidoctor-user-manual.adoc"}]]]
 :cljdoc.doc/links ["Community Slack" {:url "https://my-community.slack.com"}]}
 ```

> I found the changes you've done so far seem fine in terms of working around that, are there other places you're concerned about where this might make the code harder to maintain?

No, not right now. I think it's ok for now. I just feels like one of those features that haunt you in the future while not providing that much value (due to the alternative of adding links to your README). But probably I am just overcautious 🙂

martinklepsch:
I hadn't really thought about this but makes total sense, great idea! 🙂

rollacaster:
Yep, I like the ideas as well. The credits for that go to @rvalentini, he got the idea during our first unofficial cljdoc-contributor pub-crawl 😜

I squashed all commits since it was quite a lot back and forth.

I added a migration. The doc-tree used to be stored without the key as nested vectors. Now it's stored as a map with the `:cljdoc.doc/tree` key. I wrote a rather naive query and I am not sure if it performs well. It needs to de-/serialize all doc-trees. Is it possible to test it with the real db?

An alternative would be to check at runtime if the `:cljdoc.doc/tree` is set or not.

It's possible to nest the links, I am not sure if this is useful. The code was already there so I reused it.

martinklepsch:
> The credits for that go to @rvalentini, he got the idea during our first unofficial cljdoc-contributor pub-crawl 😜

Also that sounds fun! I'll give you a ping next time I'm around in Munich, maybe we can repeat that :)

lread:
This PR hasn't seen activity for a few years. I'm gonna go ahead and close it.

Issue #412, opened 2020-09-20
raw docstring toggle does not work from offline bundle


lread:
I don't know how many folks are using offline bundles, but I noticed that the "raw docstring" does not work for offline bundles:

![image](https://user-images.githubusercontent.com/967328/93724256-70c52300-fb73-11ea-95eb-7f6c229898a7.png)

Clicking on "raw docstring" does nothing.
If there is no good reason for it not to, it should behave as does the online version and show/hide the raw docstring.

To reproduce, download an offline bundle, ex:
```shell
curl -OJ https://cljdoc.org/download/lread/cljdoc-exerciser/1.0.34
```

lread:
This was fixed by #473, yah?

PR #473, opened 2021-09-21
Fix raw docstring toggle in offline bundles #412

rollacaster:
There were 2 causes why it was not working:
- cljdocs's `assets/js/index.js` was not included in offline bundles
- raw toggles are not initialized since the `.ns-page` css class was missing in offline bundles

Solution:
- `assets/js/index.js` is included in the offline bundle (using the same as approach as `assets/cljdoc.css` which is already included)
- Adding `.ns-page` did not work. If it is set the Scroll Indicators are initialized as well which are not available in the offline bundle. Therefore a new class `.ns-offline-page` was added which initializes the raw toggle but not the scroll indicators.

### DIFF
--- a/js/cljdoc.ts
+++ b/js/cljdoc.ts
@@ -2,6 +2,10 @@ function isNSPage(): boolean {
   return !!document.querySelector(".ns-page");
 }
 
+function isNSOfflinePage(): boolean {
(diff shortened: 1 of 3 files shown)


corasaurus-hex:
Looks great, thanks for the contribution @rollacaster!!

PR #476, opened 2021-10-04
Add previously visited doc links on homepage

### LINKED ISSUE #282: Show list of last viewed projects on homepage

rollacaster:
There's already an abandoned PR for this issue #399 so I thought I'd give it another try. I kept only the first commit of the old PR since merging the complete old PR  would be more effort.

After the page is loaded existing example links are replaced with the most recently visited doc links. If one of the initial example links is used it get's ignored.

Fixes #282

### DIFF
--- a/js/index.tsx
+++ b/js/index.tsx
@@ -13,6 +13,7 @@ import {
   toggleMetaDialog,
   addPrevNextPageKeyHandlers
 } from "./cljdoc";
+import { initRecentDocLinks } from "./recent-doc-links";
 
 export type SidebarScrollPos = { page: string; scrollTop: number };
 
@@ -54,6 +55,10 @@ if (isProjectDocumentationPage()) {
   addPrevNextPageKeyHandlers();
 }
 
+const docLinks = document.querySelector("#doc-links");
+if (docLinks) {
+  initRecentDocLinks(docLinks);
+}
 window.onbeforeunload = function () {
   var sidebar = Array.from(document.querySelectorAll(".js--main-sidebar"))[0];
   if (sidebar) {
--- a/js/recent-doc-links.tsx
+++ b/js/recent-doc-links.tsx
@@ -1,0 +1,32 @@
+import { h, render } from "preact";
+import { CljdocProject } from "./switcher";
+
+export const initRecentDocLinks = (docLinks: Element) => {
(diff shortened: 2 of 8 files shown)


martinklepsch:
Hi Thomas! Looking promising. I'd love to feature those a bit more prominently on the homepage as I think it's quite valuable and users might miss it when they just replace something that has been static for the last few years :D 

A sketch of what I could imagine this looking like:

> Read the rationale or pick up where you left off:
>
> - rum, _last viewed today_
> - m-parser, _last viewed 3d ago_
> - reitit, _last-viewed 20d ago_

Also very open to other designs/ideas, just trying to keep it simple :) I like the idea of showing the last time the docs were viewed because it implicitly tells the user what they're looking at (a reverse chronological list of the docs they've been looking at). The last-viewed data isn't currently in local storage though 😅 

Screen Shot 2021-10-04 at 21 16 02

I'm ok merging this as is but if you're down to hack on this a bit more, I hope these pointers are helpful :)

rollacaster:
Hi Martin! You're right probably no one will notice the change in the current design, I'll make it more prominent 👍 

and thanks for the invite 🙂

rollacaster:
Updated the UI + Showing last view time

image

martinklepsch (commented):
This looks *really* great! I wonder if we should just remove the "Read the rationale..." thing once you visited some documentation? What do you think?

rollacaster:
"Read the rationale..."  is now removed after at least three doc pages have been visited. 

If it's removed earlier it could be confusing, e.g. you visit cljdoc for the first time and decide to explore it using the examples. You click on the first example and then back, now the other examples are gone. (If we remove it earlier)

martinklepsch (approved):

Issue #426, opened 2020-10-31
Images are not visible in SNAPSHOT releases


BrunoBonacci:
Image links are broken in SNAPSHOT releases:

Here some examples where the source doc hasn't changed:

- https://cljdoc.org/d/com.brunobonacci/mulog/0.3.1/doc/%CE%BC-log-internals - OK
- https://cljdoc.org/d/com.brunobonacci/mulog/0.4.0-SNAPSHOT/doc/%CE%BC-log-internals - BROKEN
- https://cljdoc.org/d/com.brunobonacci/mulog/0.4.0/doc/%CE%BC-log-internals - OK
- https://cljdoc.org/d/com.brunobonacci/mulog/0.5.0-SNAPSHOT/doc/%CE%BC-log-internals - BROKEN
- https://cljdoc.org/d/com.brunobonacci/mulog/0.5.0/doc/%CE%BC-log-internals - OK

The same issue exists in all the pages and all the images.

rvalentini:
I had a look at this issue and think I found the problem.

Looking at one of the broken image links:
`https://github.com/BrunoBonacci/mulog/raw/refs/heads/master/doc/images/mulog-internals.png`

The interesting part that is actually breaking the link is `/refs/heads/master/`.

This part is taken from the `:commit` key from the `scm` map here
https://github.com/cljdoc/cljdoc/blob/1ba822fe43a92a26eae2e5a4afc0aeda8e85a6d6/src/cljdoc/util/fixref.clj#L30-L32

We are setting the `:commit` to the revision extracted during the git analysis
https://github.com/cljdoc/cljdoc/blob/1ba822fe43a92a26eae2e5a4afc0aeda8e85a6d6/src/cljdoc/server/ingest.clj#L60-L61

And the revision is initially extracted here:
https://github.com/cljdoc/cljdoc/blob/1ba822fe43a92a26eae2e5a4afc0aeda8e85a6d6/src/cljdoc/analysis/git.clj#L60-L66

As long as the revision is a SHA-1 hash, there is no problem and the resulting raw link would be working, but in case the revision is defined as a "refname" (https://git-scm.com/docs/gitrevisions), it might break the link. In case the "short" branch name is used (e.g. `master`) it would also work fine, but when the revision is specified through the "full" branch name (e.g. `/refs/heads/master/`), the link breaks.

The solution here is to change the revision extraction from
```
(.getFullBranch (.getRepository repo))
```
to 
```
[...]

PR #477, opened 2021-10-04
Fix broken image links in SNAPSHOT releases (#426)

rvalentini:
This PR attempts to fix a bug in the git analysis step that causes broken image links in SNAPSHOT releases.

I added a detailed bug description as comment on the ticket: https://github.com/cljdoc/cljdoc/issues/426#issuecomment-933734109


martinklepsch (approved):

martinklepsch:
@rvalentini thanks for digging deep and figuring this out, should be live in a couple of minutes :)

PR #484, opened 2021-11-21
Introduce Cache-Control headers

### LINKED ISSUE #181: Properly set Cache-Control headers
We're setting etag headers already (#142) but Cache-Control is probably another good thing to have.

rvalentini:
This PR is intended to resolve #181

The goal is to improve the current HTTP caching by setting Cache-Control headers. For static assets, we want to enable cache-busting by making the assets immutable (file names contain content hash) and setting a long `max-age` cache directive.

Changes in this PR:
* Added a new `main.html` file which is never served, but only used as wrapper file for the static assets. Parcel picks up the `main.html` file as an entrypoint and creates the content-hashed versions of the static assets.

* New `static-resource-interceptor`, which extracts the content-hashed static assets generated by Parcel on `:enter` into a map. The map is attached to the request context and used wherever static assets are rendered. This turned out to be more invasive than I expected and affected a lot of namespaces. E.g. also initially broke the offline-bundling incl. `cli.clj` script.  

* New `cache-control-inteceptor`, which sets `maxage=31536000,immutable,public` cache directives for all static assests on `:leave`. 

* New `favicon-interceptor` That redirects the default /favicon.ico requests from browsers to the content-hashed version of the favicon. 

* Added two new parcel dependencies: 
(1) `transformer-image` is needed so Parcel can process "png" files
(2) `parcel-reporter-static-files-copy` handles static assets that should not be content-hashed (robots.txt) 

* Fix already existing Cache-Control header in badge-interceptor.


martinklepsch (commented):
Thanks for this thorough PR Riccardo! I think it's looking quite solid already :) 

There's a few things I'd like to note/suggest:

1. If we use `main.html` only for parcel I think we could also call it `parcel-entry-point.html` or something like that. This way the role of the file is a bit clearer I think?
2. It looks like files are no longer served from the classpath (i.e. `resources/`) and instead served directly from the file system. I think it might be good to stick to the classpath based approach for now, to keep the jar working properly and things like that. (I think this might only apply to the changes in `offline.clj`)
3. I wouldn't mind to not content-hash the favicon. We could just put a long cache duration on it and not content hash it. Primary motivation would be just reducing lines of code / maintenance surface. 

Also big thanks for noticing the broken header for the badges, the badges have been kind of flaky and maybe this improves things further :) 

I'll also request a review from @danielcompton as he's been contributing to the discussion and a second pair of eyes is always good :) 

Thank you!

rvalentini:
Thanks a lot for the super fast feedback :slightly_smiling_face::+1: I will start working on the points you mentioned right away. :muscle: 
 
Concerning your second suggestion I'm not sure I fully understand what you mean however. Before my changes in `offline.clj` the files where served from `/resources` (cljdoc.css) and `/resources-compiled` (the JS files), which both are on the classpath. Now after the changes the content-hashed version of the CSS file is also served from `/resources-compiled` together with the JS resources. What is your suggestion to improve things here?

src/cljdoc/render/offline.clj
```diff
@@ -171,9 +173,9 @@
         lib-page-features (->> doc-attrs (map :page-features) (apply merge))]
     (reduce
      into
-     [[["assets/cljdoc.css" (io/file (io/resource "public/cljdoc.css"))]]
-      [["assets/js/index.js" (io/file "resources-compiled/public/js/index.js")]]
-      [["assets/js/index.js.map" (io/file "resources-compiled/public/js/index.js.map")]]
+     [[["assets/cljdoc.css" (io/file (str static-resource-prefix (get static-resources "/main.css")))]]
+      [["assets/js/index.js" (io/file (str static-resource-prefix (get static-resources "/main.js")))]]
+      [["assets/js/index.js.map" (io/file (str static-resource-prefix (get static-resources "/main.js" ".map")))]]
```
martinklepsch, line 178: @rvalentini Here you removed `io/resource` and instead provided the resource directory as a process-relative path. I think we should continue to use `io/resource`, e.g.

```clj
(io/resource (str "public/out" (get static-resources "/main.css")))
````

(or something like that :) 

Duh, now that I write this I notice this is only true for the first asset. Overall I feel like it's preferable to serve from classpath as that is packaged with Jars etc. Maybe we can use `io/resource` for all these resources here?

rvalentini:
I added two commits addressing your suggestions (1) and (3) from above, as well as the suggested refactoring of the `cache-control-interceptor`.
Concerning (1): 
After renaming `main.html` to `parcel-entry-point.html`, I realized this will also cause the generated CSS and JS files, to have the prefix `parcel-entry-point.`, which I thought is not nice and shouldn't be communicated to the client. Parcel v1 also had the option to configure this explicitly via `--out-file` CLI flag, but this option disappeared in Parcel v2. The migration guide suggests to use the "Targets" concept instead, but I couldn't get that working, at least not for a `html` entry-point. Long story short, I ended up renaming to `cljdoc.html`, which is already a bit less misleading than calling the file `main.html`? Open to better suggestions here :)

I understand now what you mean concerning the resource directory in `offline.clj`, I will try to fix that too :+1:

rvalentini:
Fixed the offline-bundling in `offline.clj` with the last commit. The content-hashed resources are now loaded from the classpath.

rvalentini:
Added a small unit test for the resource map extraction and fixed another unit test that broke because of the latest release of https://github.com/clojure/math.numeric-tower (`0.0.4` -> `0.0.5`). First new version since 2014! I guess we have to fix that test again around 2028 :smile:

danielcompton (approved):
I haven't given this a really detailed look, but the shape of the code is what I was expecting to see.

rvalentini:
@danielcompton Thanks for having a look at this!

I just pulled in the latest changes and everything still looks fine to me. @martinklepsch any additional remarks here? It looks like I would have some spare time next week to work on some further improvement suggestions :)

martinklepsch:
Thanks for your patience @rvalentini. I'm happy to merge this with one request: could you take another look that the shape of the offline bundles is still the same? (i.e. download one using the old code and compare it to the new)

With the other aspects of this change we could easily fix any issues that might come up but the offline bundles are considered as an "API" of sorts since external apps like Dash.app rely on it. 

If you say those offline bundles are the same and it's a go then lets get this into prod! :)

rvalentini:
I compared the currently deployed version and my local master version concerning the offline bundling for two randomly picked libraries (`clj-kondo-2021.12.01` and `enlive-1.1.6`):
* same number of files
* identical filenames
* no visual differences when opened in the browser
* hash of cljdoc.css changed -> now minified by Parcel 

Looking at the file hashes I spotted a bug in the JS source map building: the content of the index.js.map was not written correctly. This bug is fixed now and the index.js.map content looks like expected, however I realized there is another more subtle issue there: The comment linking the minified index.js file to its source map is still 

`//# sourceMappingURL=cljdoc.9dc5370c.js.map`

which breaks the link in this case, since I used the original non-content-hashed file names for the offline bundling. I wanted to keep the filenames in the offline bundle untouched to avoid any breaking changes. Would it be ok to use the content-hashed filenames instead here?

rvalentini:
I changed it. The source map file is now the only file in the offline bundle that uses content-hashed naming. Maybe a little bit peculiar, but the source map link should work fine :+1:

martinklepsch (approved):

martinklepsch:
@rvalentini I invited you to become a member of the elusive group of people that can merge PRs on this repo. 🙂  Please merge whenever you see fit & it should get deployed to prod soon thereafter :tada: 

Congrats on carrying this all the way through & thank you for contributing to make cljdoc better for all its users :)

rvalentini:
@martinklepsch Thanks a lot for your trust, I feel honored to join the club :smiley:

martinklepsch:
🎉  this is looking fabulous @rvalentini! 

Screen Shot 2021-12-21 at 14 58 45

rvalentini:
Nice! :smile: I'm really happy to see that the changes actually had some impact :partying_face:

PR #489, opened 2022-01-19
Ugrade cljdoc web site docker container

lread:
See commits for details

### DIFF
--- a/ops/docker/Dockerfile
+++ b/ops/docker/Dockerfile
@@ -1,6 +1,6 @@
-FROM openjdk:8-jre-alpine
+FROM eclipse-temurin:17-alpine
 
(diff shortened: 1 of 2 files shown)


lread:
@martinklepsch (and/or anyone else with an interest!) would you mind having a quick review before I merge to master?

My local testing did not uncover any problems, but upgrading from JDK8 to JDK17 is not a tiny change. I did not see any drastic increase in RAM usage, so I think we should still be OK with the 2GB droplet. 

After a merge to master we could see how this behaves over a few days to gain some confidence, after which I shall follow up with more maintenance-related PRs.

martinklepsch (approved):

PR #496, opened 2022-01-22
Upgrade to nippy 3.1.1

### LINKED ISSUE #495: Upgrade Nippy and handle new thawing behaviour
## Nippy now treats thawing objects differently

The version of nippy that cljdoc uses will thaw any frozen object.
The current version of nippy, for security reasons, will not automatically unthaw any random object.

## Is cljdoc vulnerable?

No, but I had to carefully review how we use nippy.

This resulted in me adding [some notes on the cljdoc database](https://github.com/cljdoc/cljdoc/blob/master/doc/cljdoc-developer-technical-guide.adoc#database-overview).

We don't thaw any objects from external sources.

## Observations

Our only use of nippy that would cause the current version to raise an eyebrow is that we freeze/thaw exceptions. (Stored in `error_info` column of `builds` table).

While I don't see this as a security issue, it could be considered questionable only because:
1. It ties our deserialization to a specific class which might be in a 3rd party library which could change
2. Forces us to introduce special handling to nippy thaw to allow, what it now considers a potentially unsafe object, to be thawed. Because we have a large and sporadic dev team, I'd like to take the nippy safe route just to avoid encouraging potentially unsafe patterns in the future.

## Proposal

While upgrading nippy, we migrate `builds.error_info` to `builds.error_info2` which will instead store an exception as data. We'll use Clojure's [Throwable->map](https://clojuredocs.org/clojure.core/Throwable-%3Emap) to convert the exceptions to data.

Error info is used by the cljdoc build log.
We'll upgrade build log server and render code to read/write `builds.error_info2` making `builds.error_info` obsolete (but available in case we need to rollback or re-convert the data at some later date).

The presentation of build errors to the end-user was raw exceptions but will become a Clojure map.

## Alternatives

1. Don't upgrade nippy 
   - Pros: 
     - no change, no breakage risk
   - Cons:
      - the version of nippy we are currently using has a CVE, and although it does not directly affect cljdoc, I'd rather move to a version without a CVE to avoid any suspicions
      - cljdoc is maintained by a sporadic team of developers. I find that keeping libraries up to date makes cljdoc easier to maintain by such a team.

2. Upgrade nippy but don't migrate error_info
   - Pros:
     - less change, less breakage risk
   - Cons:
     - see observations section above.

## Related Issues

Issue #488 asks some questions about the appropriateness of the build log error.
This is out of scope for the work we are doing here.
We'll have the data and can make decisions about what to display as part of #488.

## Technical Impacts

We use [ragtime](https://github.com/weavejester/ragtime) to handle database migrations.
By default, it handles only SQL migrations, but the nippy error info migration requires some Clojure code.
We'll add support for Clojure-based migrations via [ragtime-clj](https://github.com/mariusz-jachimowicz-83/ragtime-clj).

## Next Steps
Barring any feedback/objections, I'll follow up with a PR that matches the proposal above sometime soon.

lread:
See #495 for details.

Tested against backups of cljdoc production databases.
Verified that all nippy frozen data still thaws correctly
after upgrade.

Note that build log used to render exception, then, if present
any ex-data. Since ex-data is included in the map representing
the exception, we now just show only the map.

Closes #495


lread:
I will likely merge this to main sometime tomorrow (23-Jan-2022 EST).
Happy to hear any feedback before then!

PR #512, opened 2022-02-07
Quiet pedestal broken pipe errors

### LINKED ISSUE #367: Implement catch-all to avoid Pedestal "stylobate" catch-all

lread:
If a user clicks on x and then y before x is fully delivered,
pedestal logs an extremely verbose broken pipe error.

We work around this with a strategy tonsky blogged about here:
https://tonsky.me/blog/pedestal/#running-the-app

We now simple log an info level "broken pipe" message when
this normal event occurs.

We can do something more proper after this pedestal issue is
addressed: https://github.com/pedestal/pedestal/issues/623

Closes #367

### DIFF
--- a/src/cljdoc/server/pedestal.clj
+++ b/src/cljdoc/server/pedestal.clj
@@ -33,6 +33,7 @@
             [cljdoc.util.repositories :as repos]
             [cljdoc.util.sentry :as sentry]
             [clojure.tools.logging :as log]
+            [clojure.stacktrace :as stacktrace]
             [clojure.string :as string]
             [clj-http.lite.client :as http-client]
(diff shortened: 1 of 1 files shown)


lread:
Maybe we don't need #370?

martinklepsch:
Congrats on PR#512! 😁

lread:
Thanks @martinklepsch, it was hacky but practical!

PR #516, opened 2022-02-07
maint: upgrade npm deps

lread:


lread:
Did a full build, ran under docker locally, and all seems good.

PR #525, opened 2022-02-08
ops: have jvm dump heap out of memory errors

### LINKED ISSUE #524: Dump Heap on Out of Memory Errors
## Issue
The cljdoc website is sometimes suffering out of memory errors.
It seems like these are related to outages.
It is difficult to understand what is causing these errors.

## Proposal
Have the JVM dump the heap on out-of-memory errors.
We can then have a look at the heap dump and maybe get some understanding of what is going on.

## Related Info
The cljdoc website currently runs under docker on a 2-gigabyte DigitalOcean droplet.
This droplet also hosts the load balancer.
That's not a ton of elbow room.

## Details
We can use the JVM `-XX:+HeapDumpOnOutOfMemoryError` and `-XX:HeapDumpPath` options in our docker image.  
We'll want the dump file to be accessible, so we'll likely have the JVM write it to somewhere under the cljdoc data dir.

An accidental complication to work out is that the cljdoc data dir is different when running locally and on production. 

## Next steps
I'll follow up with a PR.

lread:
Ask the JVM to dump the heap when it encounters an out
of memory error.

Apparently there is no penalty for this action.

Should be useful in diagnosing our next out of memory error.

Can analyze dump file with tools like:
- jvisualvm
- Eclipse Memory Analyzer

Closes #524

### DIFF
--- a/ops/docker/Dockerfile
+++ b/ops/docker/Dockerfile
@@ -23,4 +23,4 @@ COPY . /app
 
 EXPOSE 8000
 
-ENTRYPOINT ["clojure", "-J-Dcljdoc.host=0.0.0.0", "-M", "-m", "cljdoc.server.system"]
+ENTRYPOINT ["script/docker-entrypoint.sh"]
--- a/script/docker-entrypoint.sh
+++ b/script/docker-entrypoint.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+# cljdoc web server launch script called from docker container
+
+set -eou pipefail
+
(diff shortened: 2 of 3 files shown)


lread:
@martinklepsch, I have tested this locally and it works.
When you have a moment, can you have a quick review for any potential prod issues?

lread:
Thanks for taking a peek @martinklepsch!

> The data dir gymnastics seem a bit annoying

Yeah, an accident of history, we should have had a fixed data dir when running from within docker. But oh well!

PR #590, opened 2022-04-15
Make OpenSearch less mysterious for cljdoc devs

### LINKED ISSUE #580: dev: Support testing opensearch more easily
### Recently
I didn't know what an `opensearch.xml` file was.

### Currently 
I do now know what this file is for.
It allows our users to use cljdoc as a search engine from their web browsers, and also, in theory, do so with type-ahead suggest support.

### But...
It is a pain to test this locally.
It means tweaking URLs in `opensearch.xml` manually (the URLs cannot be relative, unfortunately).

### Proposal
At the very least, we should add some docs in the developer's guide.
- what OpenSearch is 
- how to test locally from your web browser

We should also entertain somehow getting parcel.js to optionally create a localhost-friendly version of this `opensearch.xml`. If we do this, we should consider:
- running from dev box
- running from docker locally

lread:
Now dynamically generating `opensearch.xml`, it targets:
- `http://localhost:8000` by default (this is the new bit)
- `https://cljdoc.org` when run in prod

Some folks might not know (or might have just recently learned (yeah
me)) what OpenSearch is. Added some developer docs to help out here.

Notes:
- it took a bit of convincing to get parcel.js to ignore `opensearch.xml`
- we now set content-type of `opensearch.xml` as recommended.

Closes #580


corasaurus-hex:
Nicely done! That was fast turnaround!

lread:
Thanks! I ran with your dynamic opensearch.xml idea!

PR #596, opened 2022-04-21
Sort searchset results for easier diffing

lread:
The searchset test results verifies that a pre-computed cache bundle
transforms to an expected searchset.

The cache bundle and searchset are stored in edn files.

When this verification breaks, it is difficult to see what has changed
because the searchset ordering of defs and namespaces is
non-deterministic.

It is also tricky to compare regenerated cache bundle edn because it
uses sets for defs and namespaces.

And maps are not sorted.

So to make these things easier to diff with conventional tools:
1. searchset api now returns defs and namespaces in deterministic sorted
order
2. generated edn sorts maps and sets prior to spit


corasaurus-hex (approved):
Looks good to me!!

PR #616, opened 2022-04-25
Break down docset search by section headings

### LINKED ISSUE #592: Figure out why single docset search isn't breaking down doc results by section header

lread:
No longer defaulting to generic parsing for markdown docs.
This means the lovely code that parses section headings out is now invoked.

Replaced generic doc parsing with asciidoc specific parsing.
So we now cover both our rendered document format types.

Now collecting text via JSoup nodes instead of JSoup elements.
JSoup elements skips text only nodes.

Bump the client side index version to force refetch.

Closes #592

### DIFF
--- a/src/cljdoc/render/api_searchset.clj
+++ b/src/cljdoc/render/api_searchset.clj
@@ -7,8 +7,6 @@
             [clojure.string :as str])
   (:import (org.jsoup Jsoup)))
 
-(def header-tag? #{"h1" "h2" "h3" "h4" "h5" "h6"})
-
 (defn path-for-doc
   [doc version-entity]
   (routes/url-for :artifact/doc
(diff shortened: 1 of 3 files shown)


src/cljdoc/render/api_searchset.clj
```diff
@@ -29,27 +27,119 @@
...
+  (-> section-elem .children .first .text))
+
+(defn- adoc-section-anchor [section-elem]
+  (-> section-elem .children .first .attributes (.get "id")))
+
+(defn- append-text [s elem]
+  (let [t (-> elem .text str/trim)]
+    (if (= "" t)
+      s
+      (str s t " "))))
```
corasaurus-hex, line 59: do we need a space between s and t here?

src/cljdoc/render/api_searchset.clj
```diff
@@ -29,27 +27,119 @@
...
+  (and
+   (soup-elem? elem)
+   (#{"h1" "h2" "h3" "h4" "h5" "h6"} (.tagName elem))
+   (= "a" (some-> elem .children first .tagName))))
+
+(defn- adoc-section? [elem]
+  (and (soup-elem? elem)
+       (= "div" (.tagName elem))
+       (.hasAttr elem "class")
+       (re-matches #"sect[1-9]" (.attr elem "class"))
```
corasaurus-hex, line 46: do we need anchors on this regex? currently it would match "bisect123"

corasaurus-hex (approved):
Looks great!! A couple questions but otherwise I'd merge this.

src/cljdoc/render/api_searchset.clj
lread: Thanks, for checking, I'm not sure. 

Let's say we have t1 t2 and t3.

How would s would progress...
1. ""
2. "t1 "
3. "t1 t2 "
4. "t1 t2 t3 "

So, I don't think so! But... happy to be wrong!

src/cljdoc/render/api_searchset.clj
lread: Thanks for the careful review!

I think `re-matches` must match the whole string?
If we were using `re-find` I think we'd need those anchors.

lread: Yeah, just double-checked in the REPL:

```Clojure
Clojure 1.11.1
user=> (re-matches #"sect[1-9]" "bisect123")
nil
user=> (re-matches #"sect[1-9]" "sect123")
nil
user=> (re-matches #"sect[1-9]" "sect9")
"sect9"
user=> (re-find #"sect[1-9]" "bisect123")
"sect1"
```

src/cljdoc/render/api_searchset.clj
lread: I _think_ we are fine... 
If I've got a blind spot we can fix later.

corasaurus-hex: ohh ok yes this does make sense, thanks!! 💜

PR #623, opened 2022-05-04
Turf husky

### LINKED ISSUE #622: dev: turf husky
## Currently
We use an npm tool called husky to do a git pre-commit check.
This check runs a formatting check on our front-end source.
This check is repeated by CI.

## But...
When things go wrong (for example, I have enabled gpg signing for my commits and I've not quite got it right yet) husky is just another complexity to wonder about.

And when husky's commit check fails, it can be confusing when using git from an IDE. 

## So...
I feel that husky is generally more confusing than helpful and plan to turf it.

lread:
Closes #622


corasaurus-hex (approved):
looks good to me! 🍏

PR #627, opened 2022-05-06
ci: tease out jobs, update docker images

### LINKED ISSUE #626: ci: move more lint work to lint job?
## Currently 
Our CI workflow has a "prettier" job.
It performs a JavaScript lint on our sources.

## So maybe...
Rename it to "lint" and move all lint-like work from the "build" job to this job?

- clj-kondo linting 
- cljfmt format checks

I think this makes sense. 
- more things can happen at once
- the mysterious "prettier" job gets a more meaningful name.

## Alternatively
We could just move JavaScript linting to our main "build" job.

lread:
The build job is now a noop job that depends on:
- back-end-checks - does back end linty things
  (teased out from old build job)
- front-end-checks - does front end linty things
  (was formerly prettier job)
- test - runs cljdoc tests
  (teased out from old build job)
- package - packages cljdoc for potential deployment
  (teased out from old build job)

Potential benefits are:
1. shorten build time by running jobs in parallel (TBD)
2. separting build job out allows each individual job to fail separately,
giving us more more info. (Failing slower).

Other changes:
- front-end-checks now fails build if it fails
- moved entirely away from deprecated circleci docker images and now using their new cimg images.
  (this sometimes required other changes)
- now also basing cache id on bb.edn where appropriate
- now also caching ~/.gitlibs for Clojure deps

Closes #626

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,7 +30,7 @@ jobs:
 
       - run:
           name: Bring Down Clojure deps
-          command: clojure -Spath
+          command: clojure -P -X:cli:test
 
       - run:
           name: Lint
@@ -104,7 +104,7 @@ jobs:
 
       - run:
           name: Bring Down Clojure deps
-          command: clojure -Spath
+          command: clojure -P -X:cli:test
 
       - run:
           name: Build Docs for Sample Library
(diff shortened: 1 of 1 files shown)


lread:
This PR should also turf Python warning:
```
DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7.
```
And various pip and awscli warnings about the use of old/incompatible versions.

Issue #630, opened 2022-05-16
Docset search not returing babashka.fs/which var


lread:
## Reproduction

I bring up [babashka.fs in cljdoc](https://cljdoc.org/d/babashka/fs/0.1.6/api/babashka.fs) and type `which` into the the search box.

## Expected Behavior

Search returns `babashka.fs/which`

## Actual Behavior

Search returns no result

## Diagnosis

I guessed that maybe stopwords were at play and [this seems to be the case](https://github.com/weixsong/elasticlunr.js/blob/be7fdaac67cdb3ea1e7bff3cf0bca577b02e8e84/lib/stop_word_filter.js#L185).

corasaurus-hex:
Closed via https://github.com/cljdoc/cljdoc/pull/631

PR #631, opened 2022-05-16
Try excluding var names from stop words for single docset search

corasaurus-hex:
## Problem

When a project has a var that is a stop word, that var is unfindable in single docset search.

Ex: Go to https://cljdoc.org/d/babashka/fs/0.1.5/api/babashka.fs and type "with" to find "with-temp-dir". What's missing here? Oh darn, "with" is a stop-word.

## Proposed Solution

Remove var names from stop-words for that individual index. Be sure to split on non-word characters so that we still match. Look and see if we can also match non-word characters since those are valid clojure var names (e.g. `<<`).

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,1 +1,4 @@
-{ "trailingComma": "none", "arrowParens": "avoid" }
+{
+  "trailingComma": "none",
+  "arrowParens": "avoid"
+}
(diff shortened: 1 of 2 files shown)


lread:
Heya @corasaurus-hex, sorry have been distracted. 
I think this solves what I described in #630 but what about?: a search for `with` should find the `with-temp-dir` var in babashka.fs, does it handle that too?

corasaurus-hex:
@lread this should fix it for with-temp-dir as well now! I think I would ship this where it is right now. I think we need our own custom indexing stuff in order to handle clojure symbols and such

lread:
Awesome @corasaurus-hex! If you are happy, I am too.
Also happy to review/try if you want, if not, please do make it live.

corasaurus-hex:
Thanks, Lee! 💜

lread:
Thank you @corasaurus-hex!

PR #660, opened 2022-09-02
Rework library articles tips

### LINKED ISSUE #270: Don't show articles help text if user provides a `cljdoc.edn`
If a repo contains a `cljdoc.edn` file we can assume the user knows about articles and don't need this helptext:

screenshot 2019-02-07 at 15 34 20

lread:
Formerly we were suggesting to the reader that they might want
to glance at our Articles docs. This messaging always appeared when a
library did not have more than readme and/or changelog articles.

This did not make a ton of sense and was a bit noisy/wordy.
As raised in the issue, we now assume the library author does not need
a tip if we have found their project's cljdoc.edn config.

We instead now offer a click-able "tip" in the articles header.
It is noticeable but also subtle.
When clicked, the tip shows tip content to the user.

The tip appears when:
- there is no cljdoc.edn and we've auto-discovered articles
- there are no articles because scm cannot be found
- there are no articles in a found scm

To make this work, we've move all articles under the articles header.
Formerly readme and changelog appeared above the articles header.

The new tip more effectively serves our 2 audiences:
- the library author who might want to make a change/fix
- the doc reader who will usually ignore our tip, but if curious, can
learn more.

Also: fixed vertical spacing on article toc for tocs with hierarchy.

Closes #270

### DIFF
--- a/js/cljdoc.ts
+++ b/js/cljdoc.ts
@@ -134,6 +134,20 @@ function toggleMetaDialog() {
   }
 }
 
+function toggleArticlesTip() {
+  const tipToggler = document.getElementById("js--articles-tip-toggler");
+  const tip = document.getElementById("js--articles-tip");
+  if (tipToggler && tip) {
+    tipToggler.onclick = () => {
+      if (tip.classList.contains("dn")) {
+        tip.classList.remove("dn");
+      } else {
+        tip.classList.add("dn");
+      }
+    };
+  }
+}
+
 function addPrevNextPageKeyHandlers() {
   const prevLink: HTMLAnchorElement | null = document.querySelector(
     "a#prev-article-page-link"
@@ -159,6 +173,7 @@ export {
   restoreSidebarScrollPos,
   saveSidebarScrollPos,
   toggleMetaDialog,
+  toggleArticlesTip,
   isNSPage,
   isNSOfflinePage,
   isProjectDocumentationPage,
--- a/js/cljdoc.ts
+++ b/js/cljdoc.ts
@@ -139,11 +139,7 @@ function toggleArticlesTip() {
   const tip = document.getElementById("js--articles-tip");
   if (tipToggler && tip) {
     tipToggler.onclick = () => {
-      if (tip.classList.contains("dn")) {
-        tip.classList.remove("dn");
-      } else {
-        tip.classList.add("dn");
-      }
+      tip.classList.toggle("dn");
     };
   }
 }
--- a/src/cljdoc/render/sidebar.clj
+++ b/src/cljdoc/render/sidebar.clj
@@ -48,7 +48,11 @@
                         #(contains? #{"Readme" "Changelog"} (:title %))
                         doc-tree)
         readme-and-changelog (first split-doc-tree)
-        doc-tree-with-rest (second split-doc-tree)]
+        doc-tree-with-rest (second split-doc-tree)
+        no-articles? (not (seq doc-tree))
+        no-cljdoc-config? (not (some-> cache-bundle :version :config))
+        no-scm? (not (some-> cache-bundle :version :scm))
+        articles-tip? (or no-articles? no-cljdoc-config? no-scm?)]
     [;; Upgrade notice
      (when-let [newer-v (bundle/more-recent-version cache-bundle)]
        (current-release-notice newer-v))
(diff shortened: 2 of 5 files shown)


lread:
Some before and after comparisons:

image

image

image

image

For the last example, we could argue that the git import error should be moved to the tip, but that would also mean dealing with api errors differently - a separate issue if we get to it.

lread:
Heya @corasaurus-hex, when you find some time, I would appreciate a review.
I think it is a good start (and better than what we were doing).

js/cljdoc.ts
```diff
@@ -134,6 +134,20 @@ function toggleMetaDialog() {
...
+function toggleArticlesTip() {
+  const tipToggler = document.getElementById("js--articles-tip-toggler");
+  const tip = document.getElementById("js--articles-tip");
+  if (tipToggler && tip) {
+    tipToggler.onclick = () => {
+      if (tip.classList.contains("dn")) {
+        tip.classList.remove("dn");
+      } else {
+        tip.classList.add("dn");
+      }
```
corasaurus-hex, line 146: This could be replaced with `tip.classList.toggle("dn")` and it would be a bit clearer.

src/cljdoc/render/sidebar.clj
```diff
@@ -59,40 +63,38 @@
...
-       (seq doc-tree-with-rest)
-       [:div.mb4.js--articles
-        (layout/sidebar-title "Articles" {:separator-line? (seq readme-and-changelog)})
-        [:div.mv3 (articles/doc-tree-view version-entity doc-tree-with-rest (:doc-slug-path route-params))]]
+     [:div.mb4.js--articles
+      (layout/sidebar-title [:span "Articles"
+                             (when articles-tip?
+                               [:sup.ttl
+                                [:a#js--articles-tip-toggler {:href "#"}
+                                 [:span.link.dib.ml1.v-mid.hover-blue.ttl.silver "tip"]]])])
```
corasaurus-hex, line 71: using "tip" as the link text is alright except it doesn't look clickable as-is. could we add an underline to it? and maybe make it blue like other links? I think that would make it look like some something you can interact with.

corasaurus-hex: Another common pattern is to use info/question icons like [this](https://fontawesome.com/icons/circle-info?s=solid&f=classic) or [this](https://fontawesome.com/icons/circle-question?s=regular&f=classic) instead of text, although maybe since this is a sort of warning it would make sense to use a [bang](https://fontawesome.com/icons/circle-exclamation?s=solid&f=classic)? That typically tells people there's something to interact with there. An underline + blue text would be fine, too, obviously, but I wanted to give you options. 😄

corasaurus-hex (approved):
I really like the changes! I'd merge it as-is since it's a big improvement, but there are a couple opportunities to go further that you might want to do if you have time.

js/cljdoc.ts
lread: Oh thank you! will do!

src/cljdoc/render/sidebar.clj
lread: Good points, thanks!

Cljdoc's link style does not use underlines, but the blue idea is a good thought. I just didn't want it to stand out too much. One thing I did not mention is that the tip does have a hover effect.

I did spend time searching around for icons, but in the end, thought that the word "tip" was more obvious.

corasaurus-hex: Yeah, that seems to be the sticking point with these sorts of things, making it look clickable but also not the main thing to look at. Often in cases like that I put the clickable thing to the right away from the main subject of the section, but we have that strike-through-ish line to contend with.

corasaurus-hex: Oh, it might make sense to make hover on links add an underline, too. I suppose I could make a PR for that, though.

lread: I think it was a design style choice. I'm ok with no underline.

lread: Noticing some links have a subtle hover effect, could be more consistent, I suppose.

The only links I noticed with underlines seem to be links to build errors (see shots above).

lread: Oh but maybe an underline on hover? That might be nice... (maybe what you meant?)

corasaurus-hex: Ahhh I see that, I hadn't realized we had an example in the same screenshot 😂

corasaurus-hex: For the hover behavior I was thinking about how github does it: https://github.com/clojure/clojure

If you hover over the repo links at the top, or any of the links in the readme, they get underlines. It feels like a subtle confirmation of "you can click this" without cluttering up the display of text generally.

lread: Ya, sounds good to me!

corasaurus-hex: Sometimes I feel like design is just a thousand little nudges like that which work together to give people intuition. I like the way github does it there. You can contrast it with what they do in their own docs, where information is less dense and people are generally scanning, where interlinking between docs is super important, they have underlines always on: https://docs.github.com/en/get-started/onboarding/getting-started-with-your-github-account

corasaurus-hex: But then in the side where links are more dense they use different hover effects to make it feel clickable. I wonder if they A/B tested this stuff? Sounds fun :)

lread: Hmmm... interesting!

lread:
Gonna ship it!

PR #661, opened 2022-09-02
Give doc links a nice hover underline

corasaurus-hex:
Adds an underline to links within docs.

Screen Shot 2022-09-02 at 2 54 01 PM

The link on the left is being hovered over.


lread:
Nice! Ship it!

PR #668, opened 2022-09-06
library-authors: add screenshot of the rebuild link

holyjak:
BEWARE: I am not sure how image resolution works in these .adoc files and thus whether the path I used is correct, nor how to test it. Help, pls?


lread:
Thanks @holyjak!

This is an awkward compromise in the cljdoc UI.
We want to allow folks to rebuild but also don't want to draw attention to the rebuild link because we don't want folks willy-nilly initiating a heavy process.

Your path is fine.
In adoc you can move your image down under the text but keep it in the list item with a `+`, see [here](https://docs.asciidoctor.org/asciidoc/latest/lists/continuation/#list-continuation).

Issue #703, opened 2022-11-21
Normalize styling around platform differences for docstrings and arglists in var details


lread:
## Currently
Here's an example of how cljdoc will show a var with arglists and docstrings that are different for Clojure and ClojureScript

![image](https://user-images.githubusercontent.com/967328/203151049-ca865146-cb13-4fdf-a805-1bea71d691f1.png)

## I think...
That we should use consistent styling for both for a clearer presentation.

I think the arglists styling is clearer and we could adopt the same style when docstrings differ.

So we'd have no special header when docstrings are the same for all platforms, but present

CLOJURE DOCSTRING
CLOJURESCRIPT DOCSTRING

Headers when they differ.

## Why so nitpicky?
Yeah, maybe seems a bit nit-picky, but consistent styling here will make #701 styling more straightforward for me.

martinklepsch:
I'm a bit confused by the 

> but present
> 
> CLOJURE DOCSTRING
> CLOJURESCRIPT DOCSTRING
> 
> Headers when they differ.

bit. The rest of the issue read like you'd prefer to have the same hanging clj/cljs markers as for docstrings? 

Either way I agree that making this consistent is a good idea 👌

lread:
Thanks for asking about this @martinklepsch!

Yeah, instead of the little "clj" and "cljs" prefix for docstrings, I was thinking of matching the style of the arglists header.

So for the above we'd instead have:

```
fn2
CLOJURE ARGLISTS
(fn2 a)
(fn2 a b)
CLOJURESCRIPT ARGLISTS
(fn2 a)
CLOJURE DOCSTRING
some clj specific docstring
CLOJURESCRIPT DOCSTRING
some cljs specific docstring
```

I'm toying around with it a bit. The headers are a bit wordy in some cases and steal away too much attention from the content.
This might be less header-noisy:

```
fn2
CLOJURE
 (fn2 a)
 (fn2 a b)
 some clj specific docstring
CLOJURESCRIPT
 (fn2 a)
 some cljs specific docstring
```
If only the docstring differ:

```
fn2
(fn2 a)
CLOJURE
 some clj specific docstring
CLOJURESCRIPT
 some cljs specific docstring
```

If only the arglists differ:
```
fn2
CLOJURE
 (fn2 a)
 (fn2 a b)
CLOJURESCRIPT
 (fn2 a)
Some common docstring
```

But honestly, this case of differing docstrings and arglists across platforms is pretty rare.

[...]

Issue #701, opened 2022-11-20
The hickory.core namespace is returning "An exception occurred, sorry about that!"


lread:
## On Slack
@slipset reported that https://cljdoc.org/d/org.clj-commons/hickory/0.7.2/api/hickory.core was returning an error for him.

## Repro
I tried the link and saw the following rendered by cljdoc

> An exception occurred, sorry about that!

## Initial look
Currently, [cljdoc is intentionally throwing an exception when an item occurs under multiple platforms and that item does not match for all platforms (at least for protocols, anyway)](https://github.com/cljdoc/cljdoc/blob/c791f00536cb09916a20b7ae4e10468913492266/src/cljdoc/platforms.clj#L37-L38).

Hickory meets these conditions.

Its `HiccupRepresentable` protocol members are different between clj and cljs.
The clj [as-hiccup](https://github.com/clj-commons/hickory/blob/bef62d990d070868bb2181413b80c4d2e1c23fbf/src/clj/hickory/core.clj#L20-L26) supports arglists `[this]` and `[this zip-loc]`.
The cljs [as-hiccup](https://github.com/clj-commons/hickory/blob/bef62d990d070868bb2181413b80c4d2e1c23fbf/src/cljs/hickory/core.cljs#L11-L17) supports only `[this]` arglists.

## Potential Workarounds
Off the top of my head, 3 ideas:
1. Is the `HiccupRepresentable` protocol part of the hickory public API? If not, work to suppress it from being documented.
2. The hickory lib could suppress the cljs API from being analyzed via `:cljdoc/languages ["clj"]` in `doc/cljdoc.edn`. The downside is that the nuances of the cljs API would not be presented.
3. The hickory lib could ensure that protocols match across platforms (but, maybe there is a good reason they do not).

## Next up
Verify my understanding of current behaviour/impact.

Figure out what behaviour makes sense here for cljdoc.

lread:
## Diagnosis
Ok, seems to be restricted to how cljdoc handles protocols.

### Without Protocols
Given a dummy project with simple fns:

`hello_world/ns1.clj`
```Clojure
(ns hello-world.ns1)

(defn fn1 [a])

(defn fn2 "clj docstring" ([a] a) ([a b] (+ a b)))
```

And 
`hello_world/ns1.cljs`
```Clojure
(ns hello-world.ns1)

(defn fn1 [a b])

(defn fn2 "cljs docstring" [a] a)
```

[...]

PR #704, opened 2022-11-23
dont fail hard on platform differences

martinklepsch:
I've seen a lot of these exceptions being thrown and probably we shouldn't just break everything when something like `:deprecated` is different on clojure than clojurescript.

Seems that this stems from the usage of a `defalias` helper in the [affected library](https://github.com/ptaoussanis/encore/blob/89e09ac5656369578cbd30ab0283c1bd3fe80c45/src/taoensso/encore.cljc#L4979).

Screen Shot 2022-11-23 at 14 15 17

### DIFF
--- a/src/cljdoc/platforms.clj
+++ b/src/cljdoc/platforms.clj
@@ -23,22 +23,16 @@
   (get-field [this k]
     (let [ignore-nil #{:doc}
           uniq-vals (set ((if (ignore-nil k) keep map) k platforms))]
-      (cond
-        (and (= k :type) (second uniq-vals))
+      (if-not (second uniq-vals)
+        (first uniq-vals)
         ;; Current codox doesn't return consistent results for mulitmethods that are defined
         ;; by running a function, i.e. (defn x [a] (defmulti b first))
         ;; This is kind of an unusual case but there are libraries doing that and until this issue
         ;; is fixed we don't want to break docs for those libraries completetly, see precept 0.5.0-alpha for an example
         ;; https://github.com/CoNarrative/precept/blob/73113feec5bff11f5195261a81a015f882544614/src/cljc/precept/core.cljc#L356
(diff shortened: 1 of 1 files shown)


martinklepsch:
More generally this makes me wonder if we should mark some exceptions as "safe to be displayed to end users" in order to make it easier to report and diagnose issues. Could be done by a key in `ex-data` like `:show-details-on-error-page?` or something.

martinklepsch:
FYI, I requested a free plan for Sentry that would allow me to add more people to view errors on Sentry. (https://sentry.io/for/open-source/)

lread:
Thanks @martinklepsch! I too am poking around in this area for #701. 

So cljdoc currently throws if, for example, a var or ns is deprecated for cljs but not clj?

It's maybe an unusual case but also interesting for a user to know.

There's not a ton of data. We should probably just present the differences in an easy to read way? The presentation aspect seems related to #703.

martinklepsch:
> So cljdoc currently throws if, for example, a var or ns is deprecated for cljs but not clj?

Correct. In the specific case of deprecation I think we could even assume that the any non-nil value should take precedence over a `nil` value. If there's two non-nil value then maybe that still warrants an exception or something.

lread:
> Correct. In the specific case of deprecation I think we could even assume that the any non-nil value should take precedence over a nil value. If there's two non-nil value then maybe that still warrants an exception or something.

Interesting.
What do you think of just presenting the data? 
This way cljdoc does not infer any meaning.
We currently treat deprecated as a boolean for render purposes.
We could continue to do that. 
Not sure what makes sense to render... have to think about that.
Current deprecated example:
image
Maybe `deprecated` <sup>cljs</sup> (cljs would also be rendered red).
User could click through to source to see actual deprecation value.

But... maybe I am drifting from the intent of your PR.
In the short term it probably still makes sense, yeah?

lread:
Shall we merge this one @martinklepsch?
I'll be in this area soon, if we merge now, we can avoid conflicts.

PR #710, opened 2022-12-05
Indicate when a var has diffs across platforms

### LINKED ISSUE #702: Consider indicating when a var differs between platforms in var list
## Currently
Cljdoc indicates the dominant platform for a namespace and annotates all vars that differ from the dominant platform.

For example, if most functions are in defined in both Clojure and Clojure it will indicate such, and annotate each var that differs. If a var is only defined in Clojure, it will mark it as such and if a var is only defined in ClojureScript it will mark it as such.

Example presentation:

image

## But...
When a var is defined in both Clojure and ClojureScript there is no indication in the var index when the var's arglists and/or docstrings differ for those platforms.  

Note that the difference is shown when viewing var details. Example presentation when both arglists and docstrings differ:
image

This will not come up terribly often in a lib, but when it does, it is very useful information to see at-a-glance.

## Proposal
Indicate vars that differ for platforms with superscript of something like: clj!=cljs

This will sometimes override the current annotation of: clj/s (which indicates a var is defined for  both clj and cljs).

lread:
Existing var annotations are `cljs`, `cljs` and `clj/s`.

A new `clj/s≠` annotation is used for vars where there are differences in docstring, arglists and/or protocol members across platforms.

When differences exist, this new annotation is always shown in the var index regardless of dominant platforms.

Var annotations now always shown for all vars in var detail view, including protocol members.

There is no var index for offline docs, but var detail is affected. Will verify macOS Dash app is ok with this after deploy and adjust as necessary.

Closes #702

### DIFF
--- a/src/cljdoc/render.clj
+++ b/src/cljdoc/render.clj
@@ -105,7 +105,7 @@
             {:top-bar top-bar-component
              :main-sidebar-contents (sidebar/sidebar-contents route-params cache-bundle last-build)
              :vars-sidebar-contents (when (seq ns-defs)
-                                      [(api/platform-support-note platf-stats)
+                                      [(api/platforms-supported-note platf-stats)
                                        (api/definitions-list ns-emap ns-defs {:indicate-platforms-other-than dominant-platf})])
              :content (api/namespace-page {:ns-entity ns-emap
                                            :ns-data ns-data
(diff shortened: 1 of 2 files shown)


lread:
Dash seems fine. Tested by downloading cljdoc-exerciser

PR #734, opened 2022-12-24
Upgrade to highlight.js v11

lread:
I was on the fence about this one as v11 removed auto-merging of HTML into formatted blocks. This feature supports AsciiDoc callouts. See: https://docs.asciidoctor.org/asciidoc/latest/verbatim/callouts/ if you need a reminder as to what these are.

After more carefully reading over highlightjs/highlight.js#2889, I realized maybe it wouldn't be very difficult to simply re-introduce the feature. The issue even includes a complete JavaScript code example to do so.

I did my naive best to transcribe the sample JavaScript to TypeScript (which even caught a couple of minor issues).

It took me some time to figure out a way to expose the TypeScript function to our layout page. Parcel.js used to provide a --global option to support this use case, but that feature was removed.

In the end I just assigned the function to the browser window to make it available.

There is a new console warning in highlight.js v11 that is generated when formatted code contains unescaped HTML. I turned this off because the adoc callouts add... you guessed it... unescaped HTML. We also have good HTML sanitization, so I think we are good here.

### DIFF
--- a/js/hljs-merge-plugin.ts
+++ b/js/hljs-merge-plugin.ts
@@ -1,0 +1,198 @@
+// highlight.js 11 removed the 'HTML auto-merging' internal plugin
+// however this is required for AsciiDoc to insert callouts
+// fortunately the issue description shows how to re-enable it, by registering the plugin
+// See https://github.com/highlightjs/highlight.js/issues/2889
+
+// Naively transcribed to TypeScript
+var originalStream: Event[];
+
+/**
+ * @param value
+ * @returns escaped html
+ */
+function escapeHTML(value: string) {
+  return value
+    .replace(/&/g, "&amp;")
+    .replace(/</g, "&lt;")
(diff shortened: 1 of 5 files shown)


lread:
@corasaurus-hex I wonder if I've committed any especially horrible TypeScript/JavaScript sins here. :slightly_smiling_face: 

I'll go ahead and merge soon, but I am happy to adapt based on any feedback you might have.

corasaurus-hex:
That global state makes me itch but if it worked in the original plugin then it likely works here and I'm ok with it. Looks good!

corasaurus-hex:
And yes, assigning it to the global object (in the browser that's [typically the window](https://developer.mozilla.org/en-US/docs/Glossary/Global_object)) is the right way to make something globally available elsewhere.

lread:
Thanks @corasaurus-hex, much appreciated!

PR #748, opened 2023-02-09
remove superflouos segments from http uris

martinklepsch:
This fixes an issue with extra segments in the SCM URL field:

    <scm>
      <url>https://github.com/pitch-io/uix/tree/master/core</url>
      <tag>33de62e1f0425aaeb5249a589f0cc5c98a848f0f</tag>
      <connection></connection>
      <developerConnection></developerConnection>
    </scm>

Would previously attempt to clone from `git@github.com:pitch-io/uix/tree/master/core.git`

Leading to a `clone-failed` error in [builds like this one](https://cljdoc.org/builds/65965).

With this change we do not allow more than 2 segments after the domain which I think should work with all SCM services we integrate with.

### DIFF
--- a/src/cljdoc/util/scm.clj
+++ b/src/cljdoc/util/scm.clj
@@ -55,7 +55,7 @@
   [scm-url]
   (cond
     (string/starts-with? scm-url "http")
-    scm-url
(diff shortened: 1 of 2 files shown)


lread:
Nice to see you @martinklepsch! :heart:

Issue #737, opened 2022-12-25
API urls and http Accept header and 500 response


lread:
## Sentry logs
I am reacting to exceptions thrown by cljdoc via Sentry.io

## Currently
API calls can return a 500 response when a client makes a request with an unsupported `Accept` header.

## Exploration
Let's see how `api/search` behaves.

Works, `Accept` not specified
```
❮ curl -s -D - "https://cljdoc.org/api/search?q=rewrite-clj" 
HTTP/2 200 
cache-control: no-cache
content-security-policy: object-src 'none'
content-type: application/json
date: Sun, 25 Dec 2022 21:14:14 GMT
strict-transport-security: max-age=31536000; includeSubdomains
vary: Accept-Encoding
x-content-type-options: nosniff
x-download-options: noopen
x-frame-options: DENY
x-permitted-cross-domain-policies: none
x-xss-protection: 1; mode=block
content-length: 1460

{"count":8,"results":[{"artifact-id":"rewrite-clj","group-id":"rewrite-clj","blurb":"Rewrite Clojure code and edn","origin":"clojars","score":203.82979,"version":"1.1.45"},{"artifact-id":"rewrite-clj","group-id":"net.vemv","blurb":"Comment-/Whitespace-preserving rewriting of EDN documents.","origin":"clojars","score":93.833176,"version":"0.6.2"},{"artifact-id":"rewrite-cljs","group-id":"rewrite-cljs","blurb":"Comment-/Whitespace-preserving rewriting of EDN documents.","origin":"clojars","score":4.948702,"version":"0.4.5"},{"artifact-id":"conexp-clj","group-id":"conexp-clj","blurb":"A ConExp rewrite in clojure -- and so much more ...","origin":"clojars","score":2.6185662E-4,"version":"2.3.0"},{"artifact-id":"umschreiben-clj","group-id":"dev.nubank","blurb":"Rewrite utilities for refactoring clojure 
[...]

PR #756, opened 2023-03-12
Expose cache-bundle via the API

corasaurus-hex:
I'm spiking on an experiment to expose the cache-bundle via the API in order decouple the frontend and the backend. This should enable me (and others) to experiment with different frontends a little more. I'm not sure on the naming here or if there's any formatting we'd like to apply to this data before sending it back but this is the raw, simple version.

```
❯ curl 'http://localhost:8000/api/docsets/babashka/fs/0.1.6' | gron | head -10
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 30610    0 30610    0     0  1386k      0 --:--:-- --:--:-- --:--:-- 1992k
json = {};
json.defs = [];
json.defs[0] = {};
json.defs[0].arglists = [];
json.defs[0].arglists[0] = [];
json.defs[0].arglists[0][0] = "joined-paths";
json.defs[0].doc = "Splits a string joined by the OS-specific path-seperator into a vec of paths.\n";
json.defs[0].file = "babashka/fs.cljc";
json.defs[0].line = 754;
json.defs[0].name = "split-paths";
```

### DIFF
--- a/src/cljdoc/server/pedestal.clj
+++ b/src/cljdoc/server/pedestal.clj
@@ -595,11 +595,11 @@
                      :body (json/generate-string {:error "Could not find data, please request a build first"})})
                   (assoc ctx :response)))}))
 
-(def api-cache-bundle
-  "Creates an API response with a JSON representation of a cache bundle."
+(def api-docsets
+  "Creates an API response with a JSON representation of a cache bundle as a `docset`."
   (interceptor/interceptor
-   {:name ::api/cache-bundle
-    :enter (fn api-cache-bundle [{:keys [cache-bundle] :as ctx}]
+   {:name ::api/docsets
+    :enter (fn api-docsets [{:keys [cache-bundle] :as ctx}]
              (->> (if cache-bundle
                     {:status 200
                      :headers {"Content-Type" "application/json"}
@@ -639,9 +639,9 @@
                          (artifact-data-loader storage)
                          api-searchset]
 
-         :api/cache-bundle [(pom-loader cache)
-                            (artifact-data-loader storage)
-                            api-cache-bundle]
+         :api/docsets [(pom-loader cache)
+                       (artifact-data-loader storage)
+                       api-docsets]
 
          :ping          [(interceptor/interceptor {:name ::pong :enter #(pu/ok-html % "pong")})]
          :request-build [(body/body-params) request-build-validate (request-build deps)]
--- a/src/cljdoc/server/pedestal.clj
+++ b/src/cljdoc/server/pedestal.clj
@@ -595,6 +595,20 @@
                      :body (json/generate-string {:error "Could not find data, please request a build first"})})
                   (assoc ctx :response)))}))
 
+(def api-cache-bundle
+  "Creates an API response with a JSON representation of a cache bundle."
+  (interceptor/interceptor
+   {:name ::api/cache-bundle
+    :enter (fn api-cache-bundle [{:keys [cache-bundle] :as ctx}]
+             (->> (if cache-bundle
+                    {:status 200
+                     :headers {"Content-Type" "application/json"}
+                     :body (json/generate-string cache-bundle)}
+                    {:status 404
+                     :headers {"Content-Type" "application/json"}
+                     :body (json/generate-string {:error "Could not find data, please request a build first"})})
+                  (assoc ctx :response)))}))
+
 (defn route-resolver
   "Given a route name return a list of interceptors to handle requests
   to that route.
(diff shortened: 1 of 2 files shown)


lread:
TIL!: https://github.com/tomnomnom/gron

lread:
I think this is great for an experimental spike.
Clean, simple, raw.

To consider:

You'll likely find that you want to make changes (because spike!).
There is data in a cache bundle that is questionable.
After spiking for a while you'll hopefully learn what data useful/obsolete/internal/etc.
You might also learn that returning the data in a different structure might work better.

So maybe clearly mark the endpoint as experimental?
So users and our ragtag dev team clearly know that there will be breaking changes? Ideas:

- `/experimental/api/docsets/babashka/fs/0.1.6`
- `/test-labs/api/docsets/babashka/fs/0.1.6`
- `/cora-spike/api/docsets/babashka/fs/0.1.6`

I like `cora-spike` the most but I guess it doesn't lend itself to a general strategy for experimental APIs.

We could also even add something to the return payload:

```json
{"EXPERIMENTAL" "There will be breaking changes, send your feedback to <some github discussion or issue here>"}
```

But maybe the endpoint indication is enough? 

Minor inconsistency: in our API we have singular `searchset` and plural `docsets`. Is that on purpose?

lread:
Also just came to mind: cljdoc does quite a bit of work at render time.
Like correlating jar files to git files for example.
A client might appreciate this work to have been already done and just be in the returned data.
I think maybe you'll discover these kinds of things as you experiment.

corasaurus-hex:
> You'll likely find that you want to make changes (because spike!).
> There is data in a cache bundle that is questionable.
> After spiking for a while you'll hopefully learn what data useful/obsolete/internal/etc.
> You might also learn that returning the data in a different structure might work better.

Absolutely! There is already some info that's on the chopping block but I'm not 100% sure what's in there yet so I wanted to double-check what's there before removing it (`cache_bundle.version.jar` and maybe `cache_bundle.version.scm`).

> So maybe clearly mark the endpoint as experimental?

I like it. I was thinking about this last night as I drifted off to bed, that we need some kind of API versioning if we plan on exposing this to users and maintaining backwards compatibility. Since we don't want to commit to that right now I like the idea of sticking it under a path that's clearly for experiments and subject to change without notice.

I think `experiments` + a name would work well, maybe something like `/experiments/cora/api/docsets/....` -- `experiments` says it's going to be volatile and `cora` says who is running the experiments to avoid collisions, etc...

> Minor inconsistency: in our API we have singular `searchset` and plural `docsets`. Is that on purpose?

I'm a bit torn on this. We're a bit inconsistent on pluralization.

Singular endpoints:

- `searchset`
- `badge`

Plural endpoints:

- `docsets`
- `builds`
- `versions`

I went back and forth on this. It seems to me that REST endpoints are usually plural because they support both listing entities and fetching individual entities (`/api/users` and `/api/users/1234`). These endpoints, `searchset` and `docsets`, never list and only fetch individual entities (for now?).

What do you think we should do here?

> Also just came to mind: cljdoc does quite a bit of work at render time.
> Like correlating jar files to git files for example.
> A client might appreciate this work to have been already done and just be in the returned data.
> I think maybe you'll discover these kinds of things as you experiment.

Can you point me to where that happens? I'm sure I'll discover quite a bit of this and will need to do some processing eventually but I was thinking I could throw away git and jar information.

lread:
> I think `experiments` + a name would work well, maybe something like `/experiments/cora/api/docsets/....` -- `experiments` says it's going to be volatile and `cora` says who is running the experiments to avoid collisions, etc...

Ya that's real nice! Combines `cora-spike` with `experimental` ideas.

For pluralization, in endpoints it would be nice to be consistent.
I think we can rationalize that `docsets\...` might return more than 1 docset someday, so maybe fine, ya?
We know that `badge` will only ever return a single badge, so singular makes sense, maybe?
Bah. You are running an experiment so maybe not terribly important at this early stage.

> Can you point me to where that happens? I'm sure I'll discover quite a bit of this and will need to do some processing eventually but I was thinking I could throw away git and jar information.

Oh, maybe I was wrong here, I think this might happen while dynamically creating the cache-bundle. So, ignore that, I don't think it is related to this issue.

Could be another issue someday wondering why we do some dynamic work over and over for each request. But maybe it doesn't even matter if it is not causing any issues.

Getting rid of data you can't imagine a use for is probably a good strategy.
We can always add things in later without breaking the API.

Another thing that might be useful: I did look at what looked API-ish a while back and summarized in #737. See under "API URLs" and the comment that follows.

PR #808, opened 2023-08-15
maint: bump deps

lread:
Of note:
- updated to new Clojure download URLs for CI and docker
- new download URLs explicitly support grabbing latest Clojure so doing that too


lread:
hmm... circleci checks are not triggering... I'll let this sit a bit... circleci is reporting a "minor service outage" so probably that...

lread:
All seems good with circleci now

PR #815, opened 2023-08-20
Fixups for sidebar scroll pos save/restore

### LINKED ISSUE #790: Scroll to opened page in the left menu
I'm implementing support for navigating to pages (my cljdoc documentation) and some bookmark on the given page, from the command line `poly` tool, that I'm working on.

My left menu is quite long, and if I go to a page that lives at the bottom of the left menu, it would be nice if we could not just open the page and scroll to the bookmark/anchor, but also scroll to the page in the left menu (as it is now, you can't even see the selected page, you have to manually scroll to it).

The result is that you lose the context of where you are, especially when scrolling to a bookmark further down on a page, because then you don't see the heading of that page. Making the page visible in the left menu would increase the user experience.

Here is an example of the current behaviour, where `Workspace structure` isn't visible:
image

...but here it is:
image

lread:
Switched from localStorage to sessionStorage for scroll state. I think being more ephemeral here is the better way to go.

Noticed that restored scroll position was a bit off every time. Waiting for the dom to load fixed this jitter.

Now saving scroll state when a user clicks on a sidebar item. If that state is present and is for the same doc lib/version on load, then we restore the scroll pos.

Otherwise, if the selected sidebar item is out of view, we scroll to it. This supports things like bookmarks.

Noticed that style for highlighting of article and api items were different. They are now both a tachyons `b`.

Because our docset search input is effectively part of the sidebar and scrolls with the sidebar, I opted to not to do any automatic scrolling when the user selects a docset search match.

Closes #790

### DIFF
--- a/js/cljdoc.ts
+++ b/js/cljdoc.ts
@@ -1,4 +1,4 @@
-import { SidebarScrollPos } from "./index";
+import { SidebarScrollState } from "./index";
 
 function isNSOverviewPage(): boolean {
   return !!document.querySelector(".ns-overview-page");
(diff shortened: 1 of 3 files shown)


tengstrand:
Can confirm that this works now. Thanks!

lread:
Glad to hear it @tengstrand, thanks for following up!

PR #816, opened 2023-08-23
Support punctuation in single docset search terms

### LINKED ISSUE #594: Explore ways to match special characters in symbols for single docset search
Clojure allows all kinds of nice special characters like earmuffs (`*foo*`), arrows (`->`, `->>`, `~>>`), and similar. It's a shame that single docset search won't search for these. We should make it possible to do that.

### LINKED ISSUE #813: Can't search for words with - and . in them
When I search for a word that includes a `-` or `.` then it searches for both what's before the `.` or `-`.

For example if I search for "tools.deps" it will include pages that only have "tools" or "deps" in it.
If I search for "abc-123", it will include pages that matches both "abc" and "123", not just "abc-123".

It would be better if it only listed pages that exactly match the search string.

lread:
Searching a single docset is a client-side feature. It now supports searching for text that contains punctuation.

This can be helpful when searching for vars and/or namespaces which can contain punctuation.

Highlights of changes:
- strongly favour exact match of terms with punctuation but also match on subterms. For example a search for `cloure.core.test` would strongly prefer an exact match but also match on terms `clojure`, `core` and `test`.
- turn off stemming support, I think we are ok without it. We already do expansion matching (i.e. `ab` matches `abc` `abracadabra`, etc), and stemming might be less helpful for source code elements.
- turn off stop words. Stop words are normally interesting for prose, but for code, we probably don't want to exclude common words.
- turn off trimmer, we handle trimming ourselves in our new tokenizer

Closes #813
Closes #594

### DIFF
--- a/js/single-docset-search.tsx
+++ b/js/single-docset-search.tsx
@@ -8,8 +8,8 @@ import elasticlunr from "elasticlunr";
 
 const SEARCHSET_VERSION = 3;
 
-function tokenize(str?: string): string[] {
-  if (!arguments.length || str === null || str === undefined) return [];
+const tokenize = (str?: string): string[] => {
+  if (str === null || str === undefined) return [];
   // split on conventional whitespace
   const candidateTokens = str.toString().trim().toLowerCase().split(/\s+/);
   const resultingTokens: string[] = [];
@@ -31,9 +31,9 @@ function tokenize(str?: string): string[] {
     }
   });
   return resultingTokens;
-}
+};
 
-function subTokenize(tokens: string[]): string[] {
+const subTokenize = (tokens: string[]): string[] => {
   const resultingSubTokens: string[] = [];
   tokens.forEach(function (token: string) {
     // break down into subTokens, if appropriate, for example
@@ -48,7 +48,7 @@ function subTokenize(tokens: string[]): string[] {
     }
   });
   return resultingSubTokens;
-}
+};
 
 const origTokenizer = elasticlunr.tokenizer;
 
--- a/js/single-docset-search.tsx
+++ b/js/single-docset-search.tsx
@@ -12,44 +12,49 @@ const tokenize = (str?: string): string[] => {
   if (str === null || str === undefined) {
     return [];
   }
-  // split on conventional whitespace
+  // normalize and split on conventional whitespace
   const candidateTokens = str.toString().trim().toLowerCase().split(/\s+/);
-  const resultingTokens: string[] = [];
-  candidateTokens.forEach(function (candidate: string) {
-    // ignore if all non-alphanumeric and greater >= 7 characters
-    // this keeps things like *, <, >, +, ->> but turfs things like ===============
+
+  const longAllPunctuationRegex = /^[^a-z0-9]{7,}$/;
+  const standaloneCommentRegex = /^;+$/;
+  const superfluousPunctuationRegex = /^[.,]+|[.,]+$/g;
+  // strip leading and trailing periods and commas
+  // this gets rid of normal punctuation
+  // we leave in ! and ? because they can be interesting in var names
+  const trimSuperfluousPunctuation = (candidate: string) =>
+    candidate.replaceAll(superfluousPunctuationRegex, "");
+
+  return candidateTokens.reduce((tokens, candidate) => {
     if (
-      !/^[^a-z0-9]{7,}$/.test(candidate) &&
-      // ignore stand-alone comment characters
-      !/^;+$/.test(candidate)
+      // keep tokens like *, <, >, +, ->> but skip tokens like ===============
+      !longAllPunctuationRegex.test(candidate) &&
+      !standaloneCommentRegex.test(candidate)
     ) {
-      // strip leading and trailing periods and commas
-      // this gets rid of normal punctuation
-      // we leave in ! and ? because they can be interesting in var names
-      const token = candidate.replace(/^[.,]+|[.,]+$/g, "");
+      const token = trimSuperfluousPunctuation(candidate);
       if (token.length > 0) {
-        resultingTokens.push(token);
+        tokens.push(token);
       }
     }
-  });
-  return resultingTokens;
+    return tokens;
+  }, [] as string[]);
 };
 
 const subTokenize = (tokens: string[]): string[] => {
-  const resultingSubTokens: string[] = [];
-  tokens.forEach(function (token: string) {
+  const nonAlphaNumRegex = /[^a-z0-9]/;
+  const nonAlphaNumsRegex = /[^a-z0-9]+/;
+  return tokens.reduce((subTokens, token) => {
     // break down into subTokens, if appropriate, for example
     // clojure.core.test would break down to clojure core test
-    if (/[^a-z0-9]/.test(token)) {
-      const subTokens = token.split(/[^a-z0-9]+/);
+    if (nonAlphaNumRegex.test(token)) {
+      const subTokens = token.split(nonAlphaNumsRegex);
       subTokens.forEach(function (subToken: string) {
         if (subToken.length > 0) {
-          resultingSubTokens.push(subToken);
+          subTokens.push(subToken);
         }
       });
     }
-  });
-  return resultingSubTokens;
+    return subTokens;
+  }, [] as string[]);
 };
 
 const origTokenizer = elasticlunr.tokenizer;
--- a/js/single-docset-search.tsx
+++ b/js/single-docset-search.tsx
@@ -357,7 +357,6 @@ const search = (
   if (!searchIndex) {
     return;
   }
-
   // we'd like to favour exact matches, ex: clojure.tools.test
   const exactTokens = tokenize(query);
   // but also entertain sub tokens, ex: clojure tools test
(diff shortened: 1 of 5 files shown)


lread:
This is looking good to me, but I'll do another review tomorrow before merging.

I would not be offended by a review from a TypeScript pro if that TypeScript pro had some free time. Also totally understand if that TypeScript pro is busy!

corasaurus-hex:
My ears were burning, I'm taking a look now! 👀

js/single-docset-search.tsx
```diff
@@ -6,9 +6,57 @@ import cx from "classnames";
...
+    // break down into subTokens, if appropriate, for example
+    // clojure.core.test would break down to clojure core test
+    if (/[^a-z0-9]/.test(token)) {
+      const subTokens = token.split(/[^a-z0-9]+/);
+      subTokens.forEach(function (subToken: string) {
+        if (subToken.length > 0) {
+          resultingSubTokens.push(subToken);
+        }
+      });
+    }
```
corasaurus-hex, line 48: Nice one here, optimizing for the common case of tokens without non-alphas in them. At first glance I thought the first test was wasteful but no this is actually saving creating a lot of arrays.

js/single-docset-search.tsx
```diff
@@ -6,9 +6,57 @@ import cx from "classnames";
...
+}
 
-const SEARCHSET_VERSION = 2;
+const origTokenizer = elasticlunr.tokenizer;
+
+elasticlunr.tokenizer = Object.assign(function (str?: string): string[] {
+  const tokens = tokenize(str);
+  const subTokens = subTokenize(tokens);
+  return tokens.concat(subTokens);
+}, origTokenizer);
```
corasaurus-hex, line 59: You could do this all in one go, and use an arrow function as well:

```typescript
elasticlunr.tokenizer = Object.assign((str?: string): string[] => {
  const tokens = tokenize(str);
  const subTokens = subTokenize(tokens);
  return tokens.concat(subTokens);
}, elasticlunr.tokenizer);
```

js/single-docset-search.tsx
```diff
@@ -308,16 +349,57 @@ const search = (
...
+  const exactTokens = tokenize(query);
+  // but also entertain sub tokens, ex: clojure tools test
+  const subTokens = subTokenize(exactTokens);
+
+  const fieldQueries = [
+    { field: "name", boost: 10, tokens: exactTokens },
+    { field: "name", boost: 5, tokens: subTokens },
+    { field: "doc", boost: 4, tokens: exactTokens },
+    { field: "doc", boost: 2, tokens: subTokens }
+  ];
```
corasaurus-hex, line 366: 🔥

js/single-docset-search.tsx
```diff
@@ -308,16 +349,57 @@ const search = (
...
+    const fieldSearchResults: elasticlunr.SearchScores =
+      searchIndex.fieldSearch(
+        fieldQuery.tokens,
+        fieldQuery.field as keyof IndexItem,
+        searchConfig
+      );
+
+    for (const docRef in fieldSearchResults) {
+      fieldSearchResults[docRef] =
+        fieldSearchResults[docRef] * fieldQuery.boost;
```
corasaurus-hex, line 384: This could be:

```typescript
fieldSearchResults[docRef] *= fieldQuery.boost
```

js/single-docset-search.tsx
```diff
@@ -308,16 +349,57 @@ const search = (
...
+    }
+
+    for (const docRef in fieldSearchResults) {
+      if (docRef in queryResults) {
+        queryResults[docRef] += fieldSearchResults[docRef];
+      } else {
+        queryResults[docRef] = fieldSearchResults[docRef];
       }
```
corasaurus-hex, line 392: This could be:

```typescript
queryResults[docRef] ||= 0;
queryResults[docRef] += fieldSearchResults[docRef];
```

js/single-docset-search.tsx
corasaurus-hex: Aren't these already boosted because of `searchConfig` being passed in to the `searchIndex.fieldSearch`?

js/single-docset-search.tsx
```diff
@@ -308,16 +349,57 @@ const search = (
...
-    });
+    }
+  }
+  const results = [];
+  for (const docRef in queryResults) {
+    results.push({ ref: docRef, score: queryResults[docRef] });
+  }
```
corasaurus-hex, line 398:
```typescript
const results = queryResults.map(docRef => ({ ref: docRef, score: queryResults[docRef] }));
```

js/single-docset-search.tsx
```diff
@@ -308,16 +349,57 @@ const search = (
...
+
+  results.sort(function (a, b) {
+    return b.score - a.score;
+  });
```
corasaurus-hex, line 402:
```typescript
results.sort((a, b) => b.score - a.score)
```

package.json
```diff
@@ -5,7 +5,8 @@
     "lint": "prettier --check 'js/**/*.{js,jsx,ts,tsx}'",
     "dev": "parcel watch resources/public/cljdoc.html --dist-dir resources-compiled/public/out",
     "build": "parcel build resources/public/cljdoc.html --dist-dir resources-compiled/public/out",
-    "typecheck": "npx tsc --watch"
+    "typecheck": "npx tsc --watch",
+    "eslint": "eslint . --ext .ts,.tsx"
```
corasaurus-hex, line 9: 🚀

js/single-docset-search.tsx
corasaurus-hex: Also, if you wanted to squeeze every last ounce out of this you could combine the `tokenize` and `subTokenize` functions. You use them separately below, though, so probably not worth it.

corasaurus-hex (approved):
Heck yeah, this is really nice! I only have one functional question and the rest are just style nits. Thanks for doing this 💜

corasaurus-hex:
Oh, and this should also close https://github.com/cljdoc/cljdoc/issues/594 🎉

lread:
Wowza @corasaurus-hex, thanks for the _thorough_ review!
I'll pore over it with delight sometime soon!

js/single-docset-search.tsx
```diff
@@ -6,9 +6,57 @@ import cx from "classnames";
 
 import elasticlunr from "elasticlunr";
 
-elasticlunr.tokenizer.setSeperator(/[\s-.>=+\/]+/);
+const SEARCHSET_VERSION = 3;
+
+function tokenize(str?: string): string[] {
+  if (!arguments.length || str === null || str === undefined) return [];
```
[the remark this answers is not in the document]
lread: Thanks, will do!

js/single-docset-search.tsx
```diff
@@ -6,9 +6,57 @@ import cx from "classnames";
...
+      // strip leading and trailing periods and commas
+      // this gets rid of normal punctuation
+      // we leave in ! and ? because they can be interesting in var names
+      const token = candidate.replace(/^[.,]+|[.,]+$/g, "");
+      if (token.length > 0) {
+        resultingTokens.push(token);
+      }
+    }
+  });
+  return resultingTokens;
```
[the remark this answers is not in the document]
lread: Yes, like it, thanks very much!

js/single-docset-search.tsx
```diff
@@ -6,9 +6,57 @@ import cx from "classnames";
 
 import elasticlunr from "elasticlunr";
 
-elasticlunr.tokenizer.setSeperator(/[\s-.>=+\/]+/);
+const SEARCHSET_VERSION = 3;
+
+function tokenize(str?: string): string[] {
```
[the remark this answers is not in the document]
lread: Cool, can do, probably for all other functions too? Perhaps eslint could help with this kind of thing.

js/single-docset-search.tsx
```diff
@@ -6,9 +6,57 @@ import cx from "classnames";
...
+      const subTokens = token.split(/[^a-z0-9]+/);
+      subTokens.forEach(function (subToken: string) {
+        if (subToken.length > 0) {
+          resultingSubTokens.push(subToken);
+        }
+      });
+    }
+  });
+  return resultingSubTokens;
+}
```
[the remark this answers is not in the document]
lread: Yas, can do.

js/single-docset-search.tsx
lread: Yeah, I think keeping them separate is probably more maintainable.

js/single-docset-search.tsx
lread: Good question. Because the elasticlunr.js [search](https://github.com/weixsong/elasticlunr.js/blob/47100dab61cf6dc6184954cdf14ad55cc2540d4f/lib/index.js#L323-L361) did not support this kind of query I rolled my own, but also strongly based it on the existing search code [which does the extra boost](https://github.com/weixsong/elasticlunr.js/blob/47100dab61cf6dc6184954cdf14ad55cc2540d4f/lib/index.js#L342). Might be unnecessary, but conforms to current elasticlurn.js behaviour. So I think I'm ok with it.

js/single-docset-search.tsx
lread: Noice!

js/single-docset-search.tsx
lread: ya, that's leveling up!

js/single-docset-search.tsx
lread: hmmm... nice

js/single-docset-search.tsx
lread: Actually... `docRef in queryResult` means docRef is an array index, no?
Perhaps it is not too bad to leave this one as is.

js/single-docset-search.tsx
lread: When converted to this syntax TypeScript wanted the types, so I did this:

```
  results.sort((a: elasticlunr.SearchResults, b: elasticlunr.SearchResults) => b.score - a.score);
```
Which is maybe better type-wise, but more wordy.

js/single-docset-search.tsx
lread: Gonna leave this one.

js/single-docset-search.tsx
corasaurus-hex: Oh, weird, it must not know what the results array is then, or it must think that elements of results might be null/undefined.

js/single-docset-search.tsx
corasaurus-hex: Ahhh right, the `in`, it's such a tricky difference. Great catch!

corasaurus-hex:
Nicely done!!

PR #835, opened 2023-09-14
remove some noisy and not all that useful logs

martinklepsch:
Particularly the artifact loading logs are very verbose.


lread:
Yes, I always thought you needed these for some reason. Please feel free to turf!

PR #836, opened 2023-09-14
Bump memory allowance for cljdoc container in Nomad

martinklepsch:
We increased the server memory a while ago but I think we didn't increase the allocation in Nomad. This PR does that so that hopefully OOM issues occur less often. That said the permanent solution will likely involve tracking down where we're leaking memory...


lread:
I have vague memories of needing enough RAM for 2 instances of cljdoc for red/green deployment.
Will we still be OK for that with this change?

martinklepsch:
Ahh yes! you're right!

PR #840, opened 2023-09-15
ci: nit: name commands

lread:
Makes CircleCI job steps easier to read on web.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -265,11 +265,17 @@ jobs:
             bb --version
             docker --version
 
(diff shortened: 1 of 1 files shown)


lread:
Hmmm... seems like maven search api is down which seems to be affecting integration tests.

For example: this is after a very long time returning 504
https://search.maven.org/solrsearch/select?q=g:org.clojure&core=gav&rows=0

I'll retry tomorrow.

lread:
Ah yes, confirmed, from https://status.maven.org, I see:
image

PR #852, opened 2024-01-03
Exit on OutOfMemory Exceptions

martinklepsch:
Not the most elegant solution but at least this should reduce downtime while the OOM issue remains undiagnosed :) 

https://stackoverflow.com/questions/30530082/how-can-i-restart-jvm-on-outofmemoryerror-after-making-a-heap-dump

### DIFF
--- a/script/docker_entrypoint.clj
+++ b/script/docker_entrypoint.clj
@@ -48,6 +48,7 @@
     (println "Launching cljdoc server")
     (process/exec "clojure"
                   "-J-Dcljdoc.host=0.0.0.0"
+                  "-J-XX:+ExitOnOutOfMemoryError"
                   "-J-XX:+HeapDumpOnOutOfMemoryError"
                   (format "-J-XX:HeapDumpPath=%s" (str heap-dump-file))
                   "-M" "-m" "cljdoc.server.system")))


lread:
Seems like a great idea @martinklepsch!

We don’t suffer this issue very often, but it would be nice to not have to manually deal with it when it does occur.

Question: Will the infrastructure automatically restart cljdoc after the crash?

And I think giving up after the first oom will probably help with diagnosing the cause(s).

script/docker_entrypoint.clj
```diff
@@ -48,6 +48,7 @@
     (println "Launching cljdoc server")
     (process/exec "clojure"
                   "-J-Dcljdoc.host=0.0.0.0"
+                  "-J-XX:+CrashOnOutOfMemoryError"
```
corasaurus-hex, line 51: Will this result in the container restarting? Will we still be able to get the heap dump if that's the case? And if that's the case and it keeps crashing, will the heap dumps fill up the disk?

lread: Heap dumps should still be saved. I’ve set things up to save and compress the last 2 dumps.

martinklepsch:
> Also will we be notified when this happens? And are our databases crash-safe?

We won't be notified, and after reading a bit more we should probably use ExitOn... vs CrashOn.. as the former calls finalizers and other shutdown procedures.

lread:
> Also will we be notified when this happens? 

I think we should still see the exception on sentry

> And are our databases crash-safe? I believe sqlite is but I have no idea about solr/lucene/whatever-the-actual-indexing-technology-is-called

Good question about Lucene. 
Probably not? 
But I’ve not seen a corruption yet and this db is automatically rebuilt if deleted.

martinklepsch:
The Nomad docs indicate that jobs will be restarted: https://developer.hashicorp.com/nomad/tutorials/job-failure-handling/failures-restart

script/docker_entrypoint.clj
martinklepsch: As for the heap dumps, I think @lread configured this in a way that only the last two heap dumps are kept.

corasaurus-hex:
I wonder if we're hitting OOM if finalizers and shutdown procedures can even run. I guess we just try it and find out!

corasaurus-hex (approved):
Ship it! 🚀

lread (approved):

corasaurus-hex:
Maybe we can add it to the entrypoint? We'd have to stop using `exec` in that case, and we'd want to move to something lighter than clj for starting the subprocess. It feels like a pretty big lift. I wonder if Nomad doesn't have something for monitoring OOM?

corasaurus-hex:
As a related aside, maybe it's worth applying a [datadog open source account](https://www.datadoghq.com/partner/open-source/) so we [can monitor nomad](https://developer.hashicorp.com/nomad/docs/operations/monitoring-nomad) for these sorts of things.

lread:
> Maybe we can add it to the entrypoint? 

Ya, but I think this is a good start as is.

> We'd have to stop using `exec` in that case, and we'd want to move to something lighter than clj for starting the subprocess.

We are using bb, I think! 

> I wonder if Nomad doesn't have something for monitoring OOM?

Dunno about that, @martinklepsch might!

lread:
> As a related aside, maybe it's worth applying a [datadog open source account](https://www.datadoghq.com/partner/open-source/) so we [can monitor nomad](https://developer.hashicorp.com/nomad/docs/operations/monitoring-nomad) for these sorts of things.

Ya definitely a separate issue but visibility into logs from something like new relic or datadog would be very helpful.

corasaurus-hex:
ahh ok, `bb` is lighter, but we're currently using `process/exec` which is *nix's `exec` which replaces the currently running process with whatever you're executing, saving on memory. If we wanted the entrypoint to monitor `clj` and notify sentry on a crash then we couldn't use `exec` anymore -- we want `bb` to keep running so that it can notify. But like I said, it's a much larger lift.

lread:
> ahh ok, `bb` is lighter, but we're currently using `process/exec` which is *nix's `exec` which replaces the currently running process with whatever you're executing, saving on memory. If we wanted the entrypoint to monitor `clj` and notify sentry on a crash then we couldn't use `exec` anymore -- we would want `bb` to keep running so that it can notify. But like I said, it's a much larger lift.

I switched us to bb because bash was hurting my brain! 😃 

Alternative idea: If we can hit sentry from the entry point script, we could do so when it starts up. when we see there is a new heap dump. But I agree that this is more complicated and does not need to happen in this PR.

martinklepsch:
Definitely some downsides to making us feel this pain less 😅

I'll merge this for now and try to make some time to look at the heap dumps. :)

PR #854, opened 2024-01-09
try OOM exit in sentry interceptor

martinklepsch:
I think the reason the approach in #852 didn't work is that we're catching the OOM in the Sentry handler. 

Doing it like in this PR should also allow us to send a message to Sentry whenever the server is restarted to an OOM. 

I also had a look at the current 2 heapdumps but couldn't find anything too interesting. I wonder if collecting a few heapdumps over time could help (e.g. every 8 hours or so), apparently there's tools to diff them but I'm not too familiar. 

This PR also includes a change to not log every request, removing a lot of noise from the logs...

### DIFF
--- a/script/docker_entrypoint.clj
+++ b/script/docker_entrypoint.clj
@@ -48,7 +48,6 @@
     (println "Launching cljdoc server")
     (process/exec "clojure"
                   "-J-Dcljdoc.host=0.0.0.0"
-                  "-J-XX:+ExitOnOutOfMemoryError"
                   "-J-XX:+HeapDumpOnOutOfMemoryError"
                   (format "-J-XX:HeapDumpPath=%s" (str heap-dump-file))
                   "-M" "-m" "cljdoc.server.system")))
--- a/src/cljdoc/server/pedestal.clj
+++ b/src/cljdoc/server/pedestal.clj
@@ -701,6 +701,7 @@
          ::http/secure-headers {:content-security-policy-settings {:object-src "'none'"}}
          ::http/resource-path "public/out"
          ::http/not-found-interceptor not-found-interceptor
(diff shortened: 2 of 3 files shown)


lread:
Hi @martinklepsch! 

Some thoughts thumbed from my phone:

1. from some googling, that exit option we were trying might be buggy and maybe does not always work? 
2. trying to do some work when out of heap can be tricky because that work likely needs some heap
3. oom might be triggered by some request, so disabling per request logging might hinder diagnosing?
4. can you contrive a local adhoc test to verify this all works? maybe for some specific request temporarily add some code to consume the heap?

martinklepsch:
> 1. from some googling, that exit option we were trying might be buggy and maybe does not always work?

I think it should work but only if the OOM isn't caught. Since we run a web server that catches any exception it won't trigger the JVM option. I considered rethrowing but then also thought: "hey, maybe we can use this to add a message to Sentry!". 

> 2. trying to do some work when out of heap can be tricky because that work likely needs some heap

That's a great point, I didn't really consider this when adding the Sentry reporting. I'd maybe still give this a go? 

> 3. oom might be triggered by some request, so disabling per request logging might hinder diagnosing?

Maybe? I wonder if anyone will ever look at those request logs though :P  My hunch on this is that most of the code and behavior is similar for most requests so the OOM happens by many small cuts rather than issues with specific libraries... but who knows anyways :)  

> 4. can you contrive a local adhoc test to verify this all works? maybe for some specific request temporarily add some code to consume the heap?

Ehm, I was hoping to just test in prod but maybe you're right and I should give this a go locally :)

lread:
I chatted with @martinklepsch; since we aren't currently suffering from this symptom, I'm gonna close this one.

PR #860, opened 2024-02-24
docker: increase max jvm heap size

lread:
Our docker jvm max heap is reported as:

```
$ nomad alloc logs -stderr 488e4d6d
VM settings:
    Max. Heap Size (Estimated): 386.69M
    Using VM: OpenJDK 64-Bit Server VM
```

Our current docker container is 1600mb.
The default ram percentage used by the jvm is 25% or ~400mb. This is a conservative default.

As far as I remember, we don't have much else going on in our cljdoc docker container. So let's try bumping the ram percentage to 75% or ~1200mb.

### DIFF
--- a/script/docker_entrypoint.clj
+++ b/script/docker_entrypoint.clj
@@ -47,7 +47,8 @@
     (turf-old-heap-dumps heap-dump-dir)
     (println "Launching cljdoc server")
     (process/exec "clojure"
-                  "-J-XshowSettings:vm"
+                  "-J-XshowSettings:vm" ;; prints heap usage to to stderr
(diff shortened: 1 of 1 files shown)


lread:
Ok, cool after the change on prod I see:

```
nomad alloc logs -stderr <alloc id here>
VM settings:
    Max. Heap Size (Estimated): 1.13G
    Using VM: OpenJDK 64-Bit Server VM
```

martinklepsch:
Cool! I vaguely remember trying that before and Nomad killing the allocation when it allocated to much memory (according to what's in the job spec) but it seems like it might be working fine, which would be great!

PR #867, opened 2024-04-20
ci: switch to clojure docker image for JDK 21

lread:
The CircleCI clojure convenience image seems to not get much love anymore, it is still at JDK 17, for example.

The clojure docker image does get love, so we'll use it instead.


lread:
Doh! Clojure images do not contain node.

lread:
I'll go ahead and merge to exercise deployment jobs. It is possible that they might need a tweak.

PR #884, opened 2024-06-23
Improve link previews

martinklepsch:
This adds improved link preview images for documentation on cljdoc, powered by [Dynogee](https://dynogee.com?ref=cljdoc). Happy to provide this to cljdoc for free (would potentially ask for a backlink at some point). 

### Project default page

This is the default / first article in the doc tree, instead of showing "Readme" or "Home" it shows the description according to `pom.xml`

Other article pages show the title of the article according to the doctree

### Namespace page

Integrates namespace name, and a snippet of the ns docstring

### Slack (as an example)

In Slack these images would be displayed as below (note that I doctored that screenshot a bit, the domain, page title etc would be cljdoc's)

Screenshot 2024-06-23 at 14 25 03

PR #886, opened 2024-06-27
fix 500 error when clicking parent ns

### LINKED ISSUE #885: Selecting a parent namespace is triggering a 500 error
@NoahTheDuke reported on Slack:

> funny situation: i have a namespace that has a segment that isnt used anywhere else: lazytest.extensions.matcher-combinators exists, but lazytest.extensions doesn't. cljdoc doesn't know how to render the latter doc page: https://cljdoc.org/d/io.github.noahtheduke/lazytest/0.2.1/api/lazytest.extensions.matcher-combinators vs https://cljdoc.org/d/io.github.noahtheduke/lazytest/0.2.1/api/lazytest.extensions

I took a quick peek, and it seems to occur for any parent namespace that has no direct vars. 

Will take a look sometime soon and fix.

lread:
PR #884 did not expect ns-data to be nil for parent ns with no vars, which resulted in a throw, which resulted in a 500 error returned to cljdoc site user.

Closes #885

### DIFF
--- a/src/cljdoc/render.clj
+++ b/src/cljdoc/render.clj
@@ -140,7 +140,7 @@
          (layout/page {:title (str (:namespace ns-emap) " — " (proj/clojars-id version-entity) " " (:version version-entity))
                        :og-img-data {:id "4j9ovv5ojagy8ik"
                                      :page-title (str "(ns " (:namespace ns-emap) ")")
-                                     :subtitle (platf/get-field ns-data :doc)
+                                     :subtitle (when ns-data (platf/get-field ns-data :doc))
                                      :project-version (:version version-entity)
                                      :project-name (proj/clojars-id version-entity)}
 


NoahTheDuke:
that was so fast! thank you!

lread:
@martinklepsch I fixed a wee booboo in #884; I think I got it right, but you might want to review my change.

martinklepsch:
Ohh, sorry about that & thanks for the fix @lread & quick report @NoahTheDuke!

lread:
@martinklepsch I only check sentry.io weekly reports, so I would have noticed this... eventually! It shows 4.3k events for this one! Quite an exciting event in the normally quiet cljdoc ops world! :slightly_smiling_face:

martinklepsch:
I think cljdoc is getting crawled big time. A lot of the requests are coming from Amazonbot, not sure why they need to download every single OG image 😀

lread:
It's all to enable Alexa to answer even more questions for their customers :man_shrugging: 

https://developer.amazon.com/amazonbot

PR #895, opened 2024-07-04
don't render project version in og images

martinklepsch:
I'm realizing the version creates a bit of a combinatorial explosion for generated images... and it probably doesn't add all that much value. So to slightly reduce the number of images Dynogee generates for cljdoc I'd remove the version from the image.


lread:
You gonna merge this puppy? :slightly_smiling_face:

martinklepsch:
> You gonna merge this puppy? :slightly_smiling_face: :rocket:

Good idea 😁

PR #930, opened 2024-11-07
ops: address nomad deprecations

lread:


lread:
Blech, this failed. Entering nomad fatique!

lread:
Ah. I see my booboo. Exiting nomad fatique!

PR #945, opened 2024-11-28
fix: day suffix rendering

### LINKED ISSUE #944: minor: day suffix sometimes incorrect
This is only used on builds page, but I noticed day suffix is sometimes incorrect:

Shows: 21th, 22th, 23th
Instead of: 21st, 22nd, 23rd

lread:
Closes #944


src/cljdoc/util/datetime.clj
```diff
@@ -9,7 +9,13 @@
...
   Eg: 1st, 2nd, 3rd & 20th."
   [day]
-  (get {1 "st" 2 "nd" 3 "rd"} day "th"))
+  (if (<= 11 day 13)
+    "th"
+    (case (mod day 10)
+      1 "st"
+      2 "nd"
+      3 "rd"
+      "th")))
```
martinklepsch, line 18: haha who wrote this code 😛

lread: Hehe! I only noticed because I was looking at builds on the 23th of November!

PR #955, opened 2024-12-05
improve search result quality

martinklepsch:
this changes search results in some ways:

- searching for `uix` will now return `com.pitch/uix.dom` over `uix/dom` (the exact match does no longer outweigh popularity)
- searching for `reitit ring http server` will return some results (previously it would return none)

To explain the code changes a bit:

- boosted "exact matching" has been removed
- tokens mentioned in group/artifact ID no longer count twice (see DisjunctionMaxQuery stuff)
- popularity boosting has been changed to a bucketing approach, where all libraries with 3000+ downloads get a slight boost over the rest (the previous approach was not dealing well with outliers at the top, i.e. they would boost over other more term-relevant results)
- term frequency (tf) & inverse term frequency (idf) has been adjusted (not 100% about this but learning about it this seemed like a reasonable change)

### DIFF
--- a/src/cljdoc/server/search/search.clj
+++ b/src/cljdoc/server/search/search.clj
@@ -29,6 +29,7 @@
                              BooleanQuery$Builder
                              BoostQuery
                              DoubleValuesSource
+                             DisjunctionMaxQuery
                              IndexSearcher
                              MatchAllDocsQuery
                              Query
@@ -125,16 +126,15 @@
     (cond-> description
       (.add (text-field "blurb" (description->blurb description)))) ;; tokenized jar pom.xml blurbified description
     (add-versions versions)
-    (.add (DoubleDocValuesField. "popularity" popularity)) ;; 0 to 1 - popularity rating
+    (.add (DoubleDocValuesField. "popularity" popularity))
     ;; we'll use this as our boost score at query time
-    (.add (DoubleDocValuesField. "_score" (* 1.5 popularity)))))
+    (.add (DoubleDocValuesField. "_score" popularity))))
 
 (defn- clojars-download-boost [dl-max dl-lib]
-  (if (or (nil? dl-max) (zero? dl-max))
-    0 ;; if dl-max 0 or not available, we don't have stats yet.
-    (/ dl-lib dl-max)))
+  ;; slightly de-boost the long tail of libs that haven't been downloaded much
+  (if (> dl-lib 3000) 1 0.9))
 
-(defn- calculate-doc-popularity
+(defn- calculate-artifact-popularity-boost
   "Return document's popularity relative to other documents.
 
   For clojars this is the artifact download count divided the download count for the most downloaded artifact.
@@ -150,7 +150,7 @@
   (.updateDocument
    idx-writer
    (term "id" (artifact->index-id artifact))
-   (artifact->doc artifact (calculate-doc-popularity clojars-stats artifact))))
+   (artifact->doc artifact (calculate-artifact-popularity-boost clojars-stats artifact))))
 
 (defn- track-indexing-status [{:keys [last-report-time] :as status}]
   (let [status (update status :artifacts-indexed inc)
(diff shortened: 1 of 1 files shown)


martinklepsch:
@lread if this looks good to you I'll clean up the changes a bit more and delete the code related to exact matching

martinklepsch:
Just searching for `reitit` also gets improved results

## Before 
Screenshot 2024-12-05 at 13 11 54


## After

Screenshot 2024-12-05 at 13 11 48

lread:
Hi @martinklepsch! I was the one who got search working.
It does not mean it is the only way to go, but I thought a long time about the current strategy. The idea around the exact match boost is to make it possible for folks to actually find less popular libraries.

Some notes for current approach are in #588.

martinklepsch:
@lread he, makes perfect sense that I'd oversee a bunch of concerns here, will read up :)

lread:
Yeah, a big part of my design goal was avoid some special search language and/or punctuation and also give decent results. [If you look at clojars, it does expose a search language](https://github.com/clojars/clojars-web/wiki/Search-Query-Syntax), but I wanted to avoid that complexity.

lread:
Have not tested your change at all. Just sharing that I spent a ton of time coming up with current implementation.

martinklepsch:
> Have not tested your change at all. Just sharing that I spent a ton of time coming up with current implementation.

Fair, and I'm realizing I could've asked you about some things... but then again I was just toying around 😇 Would love to have you try it and see how it feels. I think it might address some of the issues you encountered in different ways but in my testing I was (and am) quite happy with the results.

lread:
> searching for `uix` will now return `com.pitch/uix.dom` over `uix/dom` (the exact match does no longer outweigh popularity)

So my approach, as you saw, deliberately strongly favoured exact match over popularity.
Your approach favours popularity over exact match.
Different approaches. Which is better? 🤷 

If you are happy with the new behaviour, and folks can still find libraries they are looking for, I guess, why not?

I'll try out your changes sometime soon to get a feel.

lread:
> Fair, and I'm realizing I could've asked you about some things... but then again I was just toying around 😇

Hey, yeah, true, but I probably would have biased you to not bother (with a "what problem are you trying to solve?") and we might have missed out on a potentially interesting exploration?

martinklepsch:
> So my approach, as you saw, deliberately strongly favoured exact match over popularity.
> Your approach favours popularity over exact match.

I was actually initially trying to also do away with popularity but I think you need a slight bit of signal to favor the most popular 2.5% of libs (roughly every lib with 3000 or more downloads). While this is a bit arbitrary the results seemed good. 

Regardless of popularity, the most impactful change here (I think) is the use of `DisjunctionMaxQuery`, which does a lot to avoid "double scoring" for the same word appearing in the group and the artifact ID. 

> Different approaches. Which is better? 🤷

This is probably best judged on a case by case basis 😅

lread:
Honestly, this is probably also a case of me thinking, "Ugh it was so hard to get this working, and the result was pretty darn good, and I'm not terribly interested in ever looking at this code again". But for you, Martin (and only you!) I will take a peek. ❤️

lread:
Ok, I am finally trying your change @martinklepsch. :heart: 

I strongly prefer the current design with regard to exact match. 

My use case is that I am looking for docs of a lib. I typically know the name of the lib. Or close to it. So I'd like to see exact matches first, followed by less exact matches sorted by recent popularity. (which is why I designed search the way it behaves!)

For example, if I search for fs, I see an exact match `fs`, followed by `babashka/fs`, `me.raynes/fs`, then `clj-commons/fs`, this seems great to me.

With this PR, I still see `fs` come up first, but in 2nd place is `me.raynes/fs`. I feel that is misleading; `me.raynes/fs` is trending down in popularity while `babashka/fs` is trending up.

Another example is searching for `clojure`. Top match is `org.clojure/clojure` for both, but the current implementation strongly then favors libs named `clojure`, whereas this PR does not.

Let's say I'm interested in artifacts from a group-id. In the current search, if I search for `org.babashka` I see a list of `org.babashka` libs sorted by popularity; in the new search, we entirely lose this because exact match weighting is gone.

I am happy with the current search behaviour and don't see a need for a change here.

martinklepsch:
Ok that's fair @lread — thanks for having a look anyways! 

The group ID search is definitely de-emphasized in this PR, my logic for this was that people usually search for docs of a specific project so optimizing for good matches there is important. 

Anyways, thanks for indulging in my explorations <3

lread:
Cljdoc is your baby, and you can commit/push/merge whatever you like, but as the current primary cljdoc maintainer, I very much appreciate your understanding here, @martinklepsch! :heart:

PR #963, opened 2025-01-02
Pull in rsbuild

corasaurus-hex:
Pulls in [rsbuild](https://rsbuild.dev/) to replace parcel. Almost have it working, just need to get the new filenames into `pedestal/build-static-resource-map`.


lread:
Superseded by https://github.com/cljdoc/cljdoc/pull/971
(thanks @corasaurus-hex!)

Issue #962, opened 2024-12-30
chore: bb compile-js watch no longer works


lread:
I setup `bb compile-js-watch` to support client side dev. 
It should watch for changes in client side sources and recompile.

It is currently failing with:
```
Error: ENOENT: no such file or directory, open '/home/lee/proj/oss/cljdoc/cljdoc/node_modules/@parcel/runtime-browser-hmr/lib/runtime-cccd0301d7c19893.js'

  Error: ENOENT: no such file or directory, open '/home/lee/proj/oss/cljdoc/cljdoc/node_modules/@parcel/runtime-browser-hmr/lib/runtime-cccd0301d7c19893.js'

unknown: File watch event emitted with 20 events. Sample event: [create] /home/lee/proj/oss/cljdoc/cljdoc/resources-compiled/public
@parcel/core: RequestGraph.respondToFSEvents duration: 0
```

This is part of Parcel.js I've never fully understood this complex tool.
It changes from time to time, and we figure out what do to to get it working again.

I wonder if there is a simpler alternative to parcel.js for us to use.

lread:
Maybe I'll give Vite a try. 
Preact recommends and supports it for hot reloading.

lread:
Let's capture current compiled resource sizes using parcel.js:

```
$ ls -lh resources-compiled/public/out 
total 1.3M
-rw-rw-r-- 1 lee lee  72K Dec 31 10:12 cljdoc.1245e9c7.js
-rw-rw-r-- 1 lee lee 685K Dec 31 10:12 cljdoc.1245e9c7.js.map
-rw-rw-r-- 1 lee lee  12K Dec 31 10:12 cljdoc.64a81faf.css
-rw-rw-r-- 1 lee lee  36K Dec 31 10:12 cljdoc.64a81faf.css.map
-rw-rw-r-- 1 lee lee  563 Dec 31 10:12 cljdoc.html
-rw-rw-r-- 1 lee lee 2.0K Dec 31 10:12 cljdoc-logo.4d4dc4a0.svg
-rw-rw-r-- 1 lee lee 2.7K Dec 31 10:12 cljdoc-logo-alpha.7200cb61.svg
-rw-rw-r-- 1 lee lee 2.7K Dec 31 10:12 cljdoc-logo-beta.be3ede9f.svg
-rw-rw-r-- 1 lee lee 3.2K Dec 31 10:12 cljdoc-logo-beta-square.618c8da7.png
[...]

PR #965, opened 2025-01-04
JS builds using esbuild

corasaurus-hex:
This PR does a few things:

- Upgrades to node 22 -- this is the latest LTS, no reason to not be on it
- Adds esbuild as a build tool and a shell script to run for builds
- Removes parcel

This nearly solves #962, we just need to figure out how to build that static map.


js/utils.ts
```diff
@@ -1,4 +1,4 @@
-import { SidebarScrollState } from "./index";
+import { SidebarScrollState } from "./types";
```
corasaurus-hex, line 1: There were some circular imports going on, where `cljdoc.ts` was importing types from `index.tsx` and `index.tsx` was importing a bunch of functions from `cljdoc.ts`.

js/cljdoc.tsx
```diff
@@ -15,14 +15,12 @@ import {
   toggleArticlesTip,
```
corasaurus-hex, line 15: Had to rename this file and `cljdoc.ts` to get the esbuild file name outputs to line up.

src/cljdoc/server/pedestal.clj
```diff
@@ -486,7 +486,7 @@
                ctx))}))
 
 (def build-static-resource-map
-  "Extracts all static resource names (content-hashed by Parcel) from the cljdoc.html file.
+  "Extracts all static resource names (content-hashed by esbuild) from the cljdoc.html file.
```
corasaurus-hex, line 489: Now that we're not using parcel the html file doesn't get updated with the correct filenames, so we need to change how we make this static resource map to not rely on the html file.

lread:
Thanks @corasaurus-hex, your `esbuild` cmd looks pretty similar to what I was playing with here in my experiments.

I was missing `--target=es2022` though, so thanks for that.  Do you think that goes back far enough?

I don't use asdf so `.tools-versions` does get stale. I did upgrade CircleCI to use current node a while ago.

I won't be using bash. I write cljdoc scripts in babashka and call them via bb tasks.

I have not played with the idea yet, but I was going to do the watching and manifest generation from babashka. Since `cljdoc.html` is effectively only used as a parcel.js build script, I was going to toss it.

I appreciate your help here.
Is your goal to a) provide tips/ideas/help-explore, or are you b) interested in fully implementing this?
Either is great with me, but it is good for me to know so that I can adjust my efforts accordingly.

corasaurus-hex:
>  your esbuild cmd looks pretty similar to what I was playing with here in my experiments.

cool!

> I was missing --target=es2022 though, so thanks for that. Do you think that goes back far enough?

How I chose the target was via [searching caniuse.com for es2022](https://caniuse.com/?search=es2022) and then clicking ["Feature support list"](https://caniuse.com/?feats=mdn-javascript_builtins_array_at,mdn-javascript_builtins_regexp_hasindices,mdn-javascript_builtins_object_hasown,mdn-javascript_builtins_error_cause,mdn-javascript_operators_await_top_level,mdn-javascript_classes_private_class_fields,mdn-javascript_classes_private_class_methods,mdn-javascript_classes_static_class_fields,mdn-javascript_classes_static_initialization_blocks) and then scrolling to the bottom to the summary. `es2022` has been supported by every major desktop browser and iOS Safari since March 2023, which seems fair to target for a developer site (Android versioning is confusing, because Chrome updates like desktop Chrome as of Android 5, and Android 7+ is the oldest in-use Android version these days, practically speaking).

> I don't use asdf so .tools-versions does get stale. I did upgrade CircleCI to use current node a while ago.

I use [mise](https://mise.jdx.dev/) these days, but it uses the same file, but I'm being a little pedantic. Glad you're keeping it upgraded!

> I won't be using bash. I write cljdoc scripts in babashka and call them via bb tasks.

Fair! This was more of a prototype, there's still more to update with it.

> I have not played with the idea yet, but I was going to do the watching and manifest generation from babashka. Since cljdoc.html is effectively only used as a parcel.js build script, I was going to toss it.

I was so hoping you'd say that! The `cljdoc.html` thing is such a weird little hack thing and I've always hated it. I wonder if we couldn't make the actual app do the building so that we can do hot reloading within the clojure app as well and reload assets in-process.

> Is your goal to a) provide tips/ideas/help-explore, or are you b) interested in fully implementing this?
> Either is great with me, but it is good for me to know so that I can adjust my efforts accordingly.

My goal is to do `a)`, which is basically where I am on cljdoc these days in general, so I want to encourage you to take things whichever way works best for you and how you'd like things done. Don't like TypeScript? Let's ditch it! Want to use shadow? Bring it on 😎.

Candidly, I haven't worked in Clojure in over two years and career-wise I don't intend on returning to it. I still enjoy making things in babashka, and I like doing a little open source contribution on principle, and I care about this project, but I recognize that I'm not the best person to steer it in any way. So do what's best for Lee, okay? I'm happy to help how I can!

lread:
> Candidly, I haven't worked in Clojure in over two years and career-wise I don't intend on returning to it. 

Sad for the Clojure community, but hopefully all good for you! 

It's always lovely to chat/interact, @corasaurus-hex! Thanks for your expertise and input here.

I'll keep your PRs as a handy reference while I work something out!

corasaurus-hex:
I had a half written script to do some of the work manually in babashka and I realized I deleted it 🤦🏻‍♀️ -- but I know I used [this example](https://github.com/babashka/babashka/blob/master/examples/README.md#cryptographic-hash) for hashing. So, maybe that's helpful?

lread:
Superceded by #971 
(thanks @corasaurus-hex!)

PR #971, opened 2025-01-07
Fix & simplify client-side build

lread:
Turf parcel.js and all of its mysteries and use a simple bb script with esbuild (with its much lessor mysteries!).

Thanks to @corasaurus-hex for her hints, tips and ideas.

Also reviewed client assets and turfed/optimized/modified as needed.


resources/public/tachyons.css
```diff
@@ -576,7 +576,7 @@ img { max-width: 100%; }
    Ref: http://nicolasgallagher.com/micro-clearfix-hack/ */
 .cf:before, .cf:after { content: " "; display: table; }
 .cf:after { clear: both; }
-.cf { *zoom: 1; }
+.cf { zoom: 1; }
```
coracuity, line 579: I looked into this weird/invalid CSS syntax when I was tinkering with this stuff and it turns out that the * was apparently a hack to fix something in IE6. that terrible browser is still haunting us 😭

coracuity (commented):
not very often that you get to remove 2k lines and end up with anymore consistent and reliable result with the same functionality, this is sooooo goooooood

resources/public/tachyons.css
lread: Yeah I researched and found the same! I guess since Tachyons CSS is on the dormant side, these kinds of things persist. Felt it pretty safe to turf!

lread:
> not very often that you get to remove 2k lines and end up with a more consistent and reliable result with the same functionality, this is sooooo goooooood

Well arentcha kind to notice and say so! :purple_heart:

lread:
I am adding some images, and this is working like a charm. 
I was cautious about doing this kind of thing before because I never really understood what gymnastics I would have to go through with parcel.js, but now it is easy peasy lemon squeezy!

PR #982, opened 2025-01-20
ops: human friendly durations for nomad job spec

lread:
Replace good `#nomad/seconds 1800` with even better `#nomad/duration 30m` for at-a-glance human-friendly durations.

Can express multiple units, e.g. 3m30s and decimal values, e.g 3.5m.


lread:
Yeah, kind cool, eh? Got the idea from your existing reader tags here!

PR #1012, opened 2025-03-29
Try out pedestal 0.8

lread:
This is first alpha release of pedestal 0.8. So maybe premature to push to production?

But interesting none-the-less that we needed so few changes!


lread:
In beta2, `io.pedestal.test` has been deprecated.
The new-school way is `io.pedestal.connector.test`, but `response-for` now works on a connector rather than a service-fn.

lread:
That pedestal connectors rework was not entirely obvious to me, perhaps because cljdoc generates routes dynamically and the higher level pedestal APIs (as far as I can tell) don't cater to this use case.

lread:
I've worked around this pedestal issue: https://github.com/pedestal/pedestal/issues/947, probably best to wait for a fix in a future pedestal release before merging to master.

lread:
Ok, I don't see any issues with pedestal v0.8.0-beta3. 
And rc1 will likely be mostly doc updates, so let's merge this puppy and see what happens.

PR #1034, opened 2025-05-31
vulnerability scan: suppress CVE-2024-22871

lread:
This should not be popping up, we use a later version of Clojure.


lread:
See https://github.com/dependency-check/DependencyCheck/issues/7695

PR #1201, opened 2026-08-20
maint: bump traefik

lread:
We last updated in Jan 2026.
Reviewed changelog https://github.com/traefik/traefik/releases and migration guides https://doc.traefik.io/traefik/v3.7/migrate/v3/ and did not see anything obvious that we need to change. We shall see!


lread:
All seems good after deploy

PR #1206, opened 2026-09-10
dev: try out borkdude's bb esbuild wrapper

lread:
Some notes:
1. It is certainly much quicker than spawning out to esbuild via npx.

2. I had to modify some cljs regexes to escape some chars. I expect this is due to compiling from bb squint rather than node squint.

3. Cljdoc uses preact. There is support in squint for preact via `:jsx-runtime {:import-source "preact"}`, but this did not work for me.

4. I switched to handling preact at bundle-time via `:jsx :automatic` and some react->preact `:alias` config. I think this is effectively what I was doing before switching to the wrapper, but because I was using squint at higher level, I didn't understand these details.

5. Cljdoc uses cache busting hashes in its filenames for most of its assets. Esbuild supports this, but the esbuild wrapper does not (yet) expose this functionality. I implemented my own hashing which was simple, but this meant a bit of fixup after files were generated by esbuild.

6. I find esbuild's analyze feature a nice way to see what contributes to my bundle size. This is not yet exposed by the esbuild wrapper.

7. My cljs tests previously required `cljs.test`, but this no longer worked for me. Switching to requiring `clojure.test` fixed the issue.


lread:
2. squint fix addressed regex issue

3. is fixed in squint, but I'm ok with my work-around in 4.

5. I seeem to be fine with doing my own cache-busting. 

6. addressed by using bb esbuild's new metafile support

7. squint fix addressed `cljs.test` fail

I think I'm good to merge.

